PR #47, opened 2014-11-13
Add main Ateco category on partner

gborelli:
Each partner should have at least one and only one Ateco category set as primary.
I changed the relation between res.partner and ateco.category

### DIFF
--- a/l10n_it_ateco/security/model_access_generator.py
+++ b/l10n_it_ateco/security/model_access_generator.py
@@ -1,0 +1,118 @@
+# -*- encoding: utf-8 -*-
+##############################################################################
+#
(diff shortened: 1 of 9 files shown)


l10n_it_ateco/security/model_access_generator.py
```diff
@@ -0,0 +1,118 @@
+# -*- encoding: utf-8 -*-
```
eLBati: This file should not be present I think

eLBati:
> This branch has conflicts that must be resolved

pedrobaeza:
Are you going to work on this or can it be closed?

PR #71, opened 2014-12-17
[IMP] mail_pec field in list and kanban view

archetipo:


eLBati:
@archetipo here is the PR https://github.com/archetipo/l10n-italy/pull/11

tafaRU:
:+1:  All comments have been addressed, so I proceed with the merge.

PR #106, opened 2015-04-09
Fix period for customer invoice

sergiocorato:
For supplier invoices the existing code is correct.
For customer invoices, the period has to be taken from date of invoice and after, if missing, from registration date.

### DIFF
--- a/account_invoice_entry_date/models/account.py
+++ b/account_invoice_entry_date/models/account.py
@@ -62,10 +62,16 @@ class AccountInvoice(models.Model):
                     raise Warning(_("The invoice date cannot be later than the"
                                     " date of registration!"))
 
-            date_start = inv.registration_date or inv.date_invoice \
-                or time.strftime('%Y-%m-%d')
-            date_stop = inv.registration_date or inv.date_invoice \
-                or time.strftime('%Y-%m-%d')
+            if inv.type in ['in_invoice', 'in_refund']:
+                date_start = inv.registration_date or inv.date_invoice \
+                    or time.strftime('%Y-%m-%d')
+                date_stop = inv.registration_date or inv.date_invoice \
+                    or time.strftime('%Y-%m-%d')
+            elif inv.type in ['out_invoice', 'out_refund']:
+                date_start = inv.date_invoice or inv.registration_date \
+                    or time.strftime('%Y-%m-%d')
+                date_stop = inv.date_invoice or inv.registration_date \
+                    or time.strftime('%Y-%m-%d')
 
             period_ids = self.pool.get('account.period').search(
                 cr, uid,
--- a/account_invoice_entry_date/models/account.py
+++ b/account_invoice_entry_date/models/account.py
@@ -51,9 +51,10 @@ class AccountInvoice(models.Model):
         for inv in self.browse(cr, uid, ids):
             date_invoice = inv.date_invoice
             reg_date = inv.registration_date
+            today = time.strftime('%Y-%m-%d')
             if not inv.registration_date:
                 if not inv.date_invoice:
-                    reg_date = time.strftime('%Y-%m-%d')
+                    reg_date = today
                 else:
                     reg_date = inv.date_invoice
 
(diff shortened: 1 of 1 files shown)


eLBati:
Hello @sergiocorato 
I don't think ```inv.date_invoice``` can be empty after calling
```
super(AccountInvoice, self).action_move_create(
            cr, uid, ids, context=context)
```
Moreover, this would cause ```account.move``` and ```account.move.line``` to have incoherent ```date``` and ```period_id```, doesn't it?

sergiocorato:
Hello @eLBati 
I didn't meant to say that the date_invoice is null (the code is unchanged in that part).
My commit is for another reason:
- when I register a supplier invoice, VAT has to be calculated from the registration date period
- when I register a customer invoice, VAT has to be calculated from the emission date period
Anyway, it's not common that a customer invoice is registered in a different date than the emission date, so is not a so worrying bug.

eLBati:
For instance using this line
```
date_start = inv.date_invoice or inv.registration_date or time.strftime('%Y-%m-%d')
```
implies that, as ```inv.date_invoice``` can't be empty, ```date_start``` will always be = ```inv.date_invoice```.
Thus, if ```inv.registration_date``` is not empty, ```period_id``` is retrieved using ```date_invoice```, but ```mov_date``` using ```reg_date```, that is = ```inv.registration_date```.

So, period_id and date of account.move would be incoherent.

sergiocorato:
Perhaps our VAT law is incoherent, but we must comply to it.

account_invoice_entry_date/models/account.py
```diff
@@ -62,10 +62,16 @@ def action_move_create(self, cr, uid, ids, context=None):
                     raise Warning(_("The invoice date cannot be later than the"
                                     " date of registration!"))
 
-            date_start = inv.registration_date or inv.date_invoice \
-                or time.strftime('%Y-%m-%d')
-            date_stop = inv.registration_date or inv.date_invoice \
-                or time.strftime('%Y-%m-%d')
+            if inv.type in ['in_invoice', 'in_refund']:
+                date_start = inv.registration_date or inv.date_invoice \
+                    or time.strftime('%Y-%m-%d')
```
lepistone: could you please define a variable `today = time.strftime('%Y-%m-%d')`?

sergiocorato:
done, thanks

eLBati:
Sorry, @sergiocorato I just tried to better understand the PR but I can't see what this change solves
> the period has to be taken from date of invoice and after, if missing, from registration date

What do you mean with ```if missing```? invoice date can't be missing.
Maybe could you describe the use case this changes solves?

sergiocorato:
Hi @eLBati, the use case is from dpr 633/72 art. 1. Il contribuente deve annotare entro quindici giorni le fatture emesse, nell'ordine della loro numerazione e con riferimento **alla data della loro emissione**, in apposito registro. 
So emission date have to be taken to select period, not registration date or today, if there is. If can't be missing, I can remove "or inv.registration_date or today" completely.

eLBati:
With use case I mean something like
 - I create an invoice with date ...
 - I confirm invoice ...
Observed behavior: ...
Expected behavior: ...

sergiocorato:
I have an user which register some invoices on 1° of month with last day of previous month date, so VAT registries are wrong because invoices go in the month of registration.

eLBati:
> I have an user which register some invoices on 1° of month with last day of previous month date, so VAT registries are wrong because invoices go in the month of registration

On http://2948683-8-0-2b230f.runbot2.odoo-community.org/ just created ```SAJ/2015/006```, set ```Invoice Date``` = ```2015-06-30```, added one line and confirmed.
Entry date is correct: I am not able to reproduce the error

sergiocorato:
Sorry the misunderstanding, the problem is the period (taxes are based on period, not on dates). I created SAJ/2015/007 with registration date 6th July and emission date 30th Juin, period is wrong (7th).

eLBati:
On 6 July 2015 at 21:46, Sergio Corato <notifications@github.com> wrote:

> Sorry the misunderstanding, the problem is the period (taxes are based on
> period, not on dates). I created SAJ/2015/007 with registration date 6th
> July and emission date 30th Juin, period is wrong (7th).
>
>
> So, in this case, why do you fill the 'registration date' field?
Why don't you leave it empty?

Probably we should make the 'registration date' field visible for supplier
invoices only.

sergiocorato:
All is possible, but it's legally register a customer invoice within 15
days from the emission, so if this field is hidden it would create a
missing legal feature.

Sergio Corato
Product Manager
SimplERP srl
tel. +390498597240
fax +390498596066


2015-07-09 16:32 GMT+02:00 Lorenzo Battistini <notifications@github.com>:

eLBati:
Please answer my questions

sergiocorato:
I leave it empty always, but some users could want to do that (even if I think it's an ugly practice)

eLBati:
On 9 July 2015 at 17:28, Sergio Corato <notifications@github.com> wrote:

> I leave it empty always, but some users could want to do that (even if I
> think it's an ugly practice)
>
>
>
This is why I propose to make the 'registration date' field visible for
supplier invoices only.
I think it would make everything more simple.
What do you think?

sergiocorato:
Yes, I think it's better for a simpler usage too.


2015-07-15 16:39 GMT+02:00 Lorenzo Battistini <notifications@github.com>:

eLBati:
Here we are https://github.com/OCA/l10n-italy/pull/123

PR #122, opened 2015-07-14
[FIX] Do not raise ImportError if user doesn't use module.

tafaRU:
As suggested in https://yelizariev.github.io/odoo/development/2015/07/06/external-dependencies.html

p.s: I still haven't time to test it, any volunteers?


eLBati:
tested :+1: thanks

dcorio:
:+1:  tested

PR #158, opened 2015-11-09
[imp] l10n_it_account.internal_layout: better report header markup

simahawk:
Adjust header markup into l10n_it_account.internal_layout to ease customization from 3rd party modules.


eLBati:
Thanks :+1:

archetipo:
:+1:

PR #162, opened 2015-11-26
[FIX] not existing css was used

dcorio:
the report template was using a wrong and not existing css file

### DIFF
--- a/l10n_it_account/reports/account_reports_view.xml
+++ b/l10n_it_account/reports/account_reports_view.xml
@@ -75,42 +75,38 @@
           </div>
       </template>
 
+      <template id="minimal_layout" inherit_id="report.minimal_layout">
+        <xpath expr="//head/t/script" position="replace">
+        <script>
+            function subst() {
+                var vars = {};
+                var x = document.location.search.substring(1).split('&amp;');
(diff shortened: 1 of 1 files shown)


eLBati:
Are the changes to js related to the wrong css file?

dcorio:
@eLBati i didn't touch the js code. just indented the <template> block

eLBati:
thanks :+1:

eLBati:
Oops, just tried to print VAT registers and got
[l10n_it_vat_registries.report_registro_iva.pdf](https://github.com/OCA/l10n-italy/files/45135/l10n_it_vat_registries.report_registro_iva.pdf)
So, that CSS should be anyway included by ```l10n_it_vat_registries``` or by ```l10n_it_account``` with another name
:-1: for now

dcorio:
@eLBati thanks for the test. this is clearly a bug of the l10n_it_vat_registries modules. we should add that CSS in l10n_it_vat_registries since those classes aren't shared among the reports but vat registries specific. We then first fix l10n_it_vat_registries. cc: @DAMilani

eLBati:
No activity for more than 6 months, closing. Feel free to reopen

PR #164, opened 2015-11-27
[8.0] l10n_it_fatturapa: import guards

LeartS:


l10n_it_fatturapa/bindings/__init__.py
```diff
@@ -17,5 +17,7 @@
 #    along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
 ##############################################################################
-
-from . import fatturapa_v_1_1
+try:
+    from . import fatturapa_v_1_1
+except ImportError:
+    pass # Module is not installed, avoid init error
```
eLBati: Why this? When you use this module, file fatturapa_v_1_1.py is always present

eLBati: Oh I see: pyxb is imported several times in file fatturapa_v_1_1.py
Could you just add a comment explaining it?

eLBati:
Thanks @LeartS please also see travis: ```l10n_it_fatturapa/bindings/__init__.py:23:9: E261 at least two spaces before inline comment```

l10n_it_fatturapa/bindings/__init__.py
LeartS: Yep that's it, there are a lot of statements (not only imports) that need
to be guarded against. You are right, a comment would be nice, I'll add
one.

l10n_it_fatturapa_out/wizard/__init__.py
```diff
@@ -18,4 +18,7 @@
 #
 ##############################################################################
 
-from . import wizard_export_fatturapa
+try:
```
archetipo: Why this?

eLBati: See https://github.com/OCA/l10n-italy/pull/164/files#r46173549

archetipo: ok but is useless check here, if the  l10n_it_fatturapa module is not resolved, or not?

l10n_it_fatturapa/bindings/__init__.py
archetipo: Please add a log as written in https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#importerror

LeartS:
@eLBati @archetipo I addressed your comments

eLBati:
thanks :+1:

archetipo:
@LeartS  thanks  :+1:

PR #193, opened 2016-02-15
[SEC] l10n_it_fatturapa_out: fix unsafe eval

gurneyalex:


lepistone:
:+1: from a technical POV. It could be worth trying to use `getattr` instead, although for me it's also OK to merge like that.

Someone should test if the module still works though. Thanks!

antespi:
:+1:

eLBati:
:+1:

PR #221, opened 2016-05-05
[IMP] Show original orders if user try to create an invoice from ddt …

OpenCode:
…related to invoiced orders

For issue https://github.com/OCA/l10n-italy/issues/220

### DIFF
--- a/l10n_it_ddt/models/stock_picking_package_preparation.py
+++ b/l10n_it_ddt/models/stock_picking_package_preparation.py
@@ -154,6 +154,31 @@ class StockPickingPackagePreparation(models.Model):
                                                  date=self.date)
         self.display_name = name
 
+    @api.multi
+    def create_invoice(self):
+        # ----- Check if sale order related to ddt are invoiced. Show them.
+        invoiced_sale = [
+            picking.sale_id.id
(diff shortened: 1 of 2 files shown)


l10n_it_ddt/models/stock_picking_package_preparation.py
```diff
@@ -154,6 +154,31 @@ def _compute_display_name(self):
...
+            if picking.sale_id and picking.sale_id.invoice_ids]
+        if invoiced_sale:
+            return {
+                'type': 'ir.actions.act_window',
+                'view_type': 'form',
+                'view_mode': 'tree,form',
+                'res_model': 'sale.order',
+                'target': 'current',
+                'domain': '[("id", "in", {ids})]'.format(ids=invoiced_sale),
+                }
```
eLBati: Should't this be returned only when invoicing is not based on delivery?

OpenCode: No. This is returned when an order as an invoice yet. In this way when you start to invoicing from order, you continue to invoicing from order.

eLBati:
ok, thanks :+1:

paolovalier:
:+1:

PR #224, opened 2016-05-12
[FIX] Add fix purchase tax multicompany

hurrinico:
I try to explain the problem.
If i use multicompany and each company are from different country, when i try too install new tax_account i get "Tax code already use" error.
With this fix, if you check check_purchase_tax_multicompany field on your company you can avoid all controls.


hurrinico:
Close because this fix already exists

PR #246, opened 2016-08-24
[8.0][FIX] l10n_it_corrispettivi: fix ValueError raised when trying to add a new invoice line.

tafaRU:
This PR fixes the following error:

```
ValueError: "name 'partner_id' is not defined" while evaluating '[company_id,partner_id,type,invoice_line,currency_id]'
````

This error is raised when trying to add a new invoice line.


paolovalier:
:+1:

eLBati:
:+1:

PR #257, opened 2016-10-18
[10.0] migration of l10n_it_abicab

dcorio:
Migration of l10n_it_abicab to v10


dcorio:
Thanks for your approvals but this pr is totally wrong then i delete it :)

> Trust no one

PR #272, opened 2016-10-20
10.0 mig l10n it split payment

archetipo:


eLBati:
Moving to https://github.com/OCA/l10n-italy/pull/288

PR #280, opened 2016-11-07
[FIX] fix issue#279

CiroBoxHub:


OpenCode:
@eLBati **select** parameter in fields declaration generate the loop. if we delete **select** we solve the problem. The bug is in Odoo core.
https://github.com/OCA/OCB/blob/10.0/odoo/fields.py#L462

eLBati:
I think ```select``` can be replaced by ```index```

OpenCode:
@eLBati ok

eLBati:
@CiroBoxHub prevedi di utilizzare ```index``` o preferisci lasciare la PR così com'è?

CiroBoxHub:
@eLBati ho reinserito index per il field parent_id

andrea4ever:
@eLBati mi sento tanto un'indovino in questo momento ;-)

eLBati:
waiting for squash to merge

eLBati:
squashed

PR #302, opened 2017-02-01
8.0 - account_invoice_entry_date - Invoice sequence based on registration_date

OpenCode:
When an in invoice or refund has a registration date, the invoice sequence must be based on registration date year

### DIFF
--- a/account_invoice_entry_date/models/account.py
+++ b/account_invoice_entry_date/models/account.py
@@ -23,7 +23,7 @@
 import time
 from openerp import fields, models
 from openerp.tools.translate import _
-from openerp.exceptions import Warning
+from openerp.exceptions import Warning as UserError
 
 
 class AccountInvoice(models.Model):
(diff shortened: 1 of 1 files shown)


account_invoice_entry_date/models/account.py
```diff
@@ -41,16 +41,37 @@ class AccountInvoice(models.Model):
...
+                ('date_stop', '>=', date_stop),
+                ('company_id', '=', invoice.company_id.id),
+                ('special', '!=', True),
+                ])
+        if period_ids:
+            period_id = period_ids[0]
+        else:
+            period_id = False
+        return period_id, date_start, date_stop
+
```
eLBati: You can also use ```find``` method of ```account.period```

scigghia:
👍

account_invoice_entry_date/models/account.py
OpenCode: find return id. I need id of period and start and stop date, too

eLBati:
@OpenCode puoi fare lo squash dei 3 commit?

OpenCode:
@eLBati scusa il ritardo. Squash effettuato. Travis verde.

PR #338, opened 2017-05-31
 l10n_it_ddt ADD weight_manual

eLBati:
to force weght when different from the computed one
Ported from d3c2e74727583f48d535774943a7a41369b4f654


eLBati:
Chiudo e mi sposto su https://github.com/OCA/l10n-italy/pull/250

PR #344, opened 2017-06-21
FIX l10n_it_ricevute_bancarie while duplicating

eLBati:
File "odoo/lib/python2.7/site-packages/odoo/models.py", line 4369, in <lambda>
    @api.returns('self', lambda value: value.id)
AttributeError: 'NoneType' object has no attribute 'id'

while duplicating


hurrinico:
👍

PR #357, opened 2017-07-31
[BUG] voucher payment could not be validated if payment journal (e.g.…

jackjack82:
… bank) had 'Autopost Created Moves' flagged. Now added update_check=True variable in account.move.line.write() method.


eLBati:
@alessandrocamilli ?

PR #358, opened 2017-08-03
[FIX][8.0] missing date invoice in ctx from shipping

sergiocorato:
When creating invoice from ddt, date is not put in invoice because missing in ctx.
With this mp the bug is solved


l10n_it_ddt/wizard/ddt_create_invoice.py
```diff
@@ -92,6 +92,7 @@ def create_invoice(self):
             ctx = self.env.context.copy()
             ctx['ddt_partner_id'] = partner_id  # ddts[0].partner_invoice_id.id
             ctx['inv_type'] = 'out_invoice'
+            ctx['date_inv'] = self.date
```
eLBati: @sergiocorato Non riesco a trovare dove ```date_inv``` venga poi usato

eLBati: Trovato: addons/stock_account/stock.py

PR #359, opened 2017-08-05
[ADD] added automatic settlemet of riba lines, using fa-check button …

jackjack82:
…on riba line, and button 'Settle all' that allow to settle all the riba lines that are in state 'Accredited'


eLBati:
@pedrobaeza Hi, do you know why transifex now fails https://travis-ci.org/OCA/l10n-italy/jobs/304619990 ?

Issue #409, opened 2017-11-18
[10.0][DDT] stock.picking.package.preparation.display_name depends on itself


2zx:
warning on module load:
odoo.fields:Field stock.picking.package.preparation.display_name depends on itself; please fix its decorator @api.depends()

OpenCode:
Pr merged. We can close this issue

PR #410, opened 2017-11-18
fix display_name depends on itself

2zx:
fix odoo warning 
"odoo.fields:Field stock.picking.package.preparation.display_name depends on itself; please fix its decorator @api.depends()"

patch to issue #409 renaming compute method and filling @api.depends decorator


l10n_it_ddt/models/stock_picking_package_preparation.py
```diff
@@ -96,7 +96,7 @@ def _default_ddt_type(self):
     carrier_id = fields.Many2one(
         'res.partner', string='Carrier')
     parcels = fields.Integer('Parcels')
-    display_name = fields.Char(string='Name', compute='_compute_display_name')
+    display_name = fields.Char(string='Name', compute='_compute_clean_display_name')
```
eLBati: Please respect PEP8 line length < 80

PR #444, opened 2018-02-01
[8.0][FIX] date format in riba line created

sergiocorato:
Riba lines created from move lines have date non formatted for Italy (date is a char field). With this mp date are put correctly.

### DIFF
--- a/l10n_it_ricevute_bancarie/__openerp__.py
+++ b/l10n_it_ricevute_bancarie/__openerp__.py
@@ -26,7 +26,7 @@
 
 {
     'name': "Ricevute Bancarie",
-    'version': "8.0.1.3.0",
+    'version': "8.0.1.3.1",
     'author': "Odoo Community Association (OCA)",
     'category': "Accounting & Finance",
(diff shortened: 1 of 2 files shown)


OpenCode:
I think is better to convert field to date type so Odoo can format it based on user language

sergiocorato:
This is not possible because when the ri.ba. are grouped by partner by date due, this field is used to summarize multiple invoice dates

eLBati:
What would happen when applying this change to existing DBs?

sergiocorato:
With this MP, nothing happens to existing DBs, only new record are translated correctly. The previous records remain in the english format (field is a char).
OpenCode comment is not applicable in this case.

eLBati:
@sergiocorato ok, thanks. Could you also do this for 10.0?

eLBati:
@OpenCode mergio?

sergiocorato:
@OpenCode un attimo che la sistemo

sergiocorato:
@OpenCode  @eLBati fatto, ho aggiornato anche la versione

PR #459, opened 2018-02-28
FIX account_vat_period_end_statement tests

eLBati:
In cases like today, feb 28, no date range is found


tafaRU:
Fast tracking this

PR #460, opened 2018-03-06
[FIX] configuration in DDT type to restrict pickings to be in only one DDT

SimoRubi:
Hi,
in DDT module, pickings can be grouped in different DDTs, this fix adds a configuration in the DDT type to enable/disable this behaviour. Please check.


SimoRubi:
@eLBati could you please check?
Thanks

PR #461, opened 2018-03-07
[FIX] fiscal position field does not exist in this model

SimoRubi:
Fiscal position field does not exist in `account.invoice.withholding.tax` model, this breaks runbot builds


tafaRU:
Fast tracking this.

PR #503, opened 2018-06-01
used multi tax_id in account invoice line and print vat registries

CiroBoxHub:
print report vat registries also there are multi tax in move line, but only if only one is included on vat registries.

deduct move_line debit to total if there is a tax and it is excluded from vat registries


tafaRU:
@CiroBoxHub vedi https://github.com/OCA/l10n-italy/pull/503#issuecomment-395108610
Grazie.

eLBati:
Chiudo per inattività

PR #505, opened 2018-06-07
[IMP] Improve fiscal position management for corrispettivi

SimoRubi:
When creating a sale order for a partner having *Use corrispettivi*, it might be useful to register the transaction in particular accounts specifically dedicated to corrispettivi.

With this PR, this can be done by creating a fiscal position having the flag *Corrispettivi* enabled: this fiscal position is automatically assigned to an user having the flag *Use corrispettivi* enabled.


SimoRubi:
Still working on this

PR #545, opened 2018-07-06
[10.0 ]removed lint=check and pep8: changed digits_compute > digits and set …

jackjack82:
…name_get() for tax.kind.

Still missing, but in odoo core:
 
2018-06-22 14:42:20,147 182 WARNING openerp_test py.warnings: /.repo_requirements/odoo/odoo/models.py:362: UnicodeWarning: Unicode equal comparison failed to convert both arguments to Unicode - interpreting them as being unequal
elif not all(cols[field.name][key] == vals[key] for key in vals):

### DIFF
--- a/l10n_it_account_tax_kind/model/account_tax_kind.py
+++ b/l10n_it_account_tax_kind/model/account_tax_kind.py
@@ -7,17 +7,17 @@ from odoo import api, models, fields
 class AccountTaxKind(models.Model):
 
     _name = 'account.tax.kind'
-    _rec_name = 'display_name'
 
     code = fields.Char(string='Code', size=3, required=True)
     name = fields.Char(string='Name', required=True)
-    display_name = fields.Char(string='Name', compute='_compute_display_name')
 
-    @api.depends('code', 'name')
     @api.multi
-    def _compute_display_name(self):
-        for record in self:
-            record.display_name = u'[%s] %s' % (record.code, record.name)
+    def name_get(self):
+        res = []
+        for item in self:
+            name = "[{}] {}".format(item.code, item.name)
+            res.append((item.id, name))
+        return res
 
     @api.model
     def name_search(self, name, args=None, operator='ilike', limit=100):
--- a/l10n_it_account_tax_kind/tests/test_account_tax_kind.py
+++ b/l10n_it_account_tax_kind/tests/test_account_tax_kind.py
@@ -11,12 +11,6 @@ class TestAccountTaxKind(TransactionCase):
         super(TestAccountTaxKind, self).setUp()
         self.tax_kind_n1 = self.env.ref('l10n_it_account_tax_kind.n1')
 
-    def test_compute_display_name(self):
-        self.tax_kind_n1._compute_display_name()
-        self.assertEqual(
-            self.tax_kind_n1.display_name,
-            u'[%s] %s' % (self.tax_kind_n1.code, self.tax_kind_n1.name))
-
     def test_name_search(self):
         result = self.env['account.tax.kind'].name_search('Escluse ex art. 15')
         self.assertEqual(result and result[0][0], self.tax_kind_n1.id)
--- a/account_central_journal/account.py
+++ b/account_central_journal/account.py
@@ -30,8 +30,8 @@ class account_fiscalyear(orm.Model):
         'date_last_print': fields.date('Last printed date', readonly=True),
         'progressive_page_number': fields.integer('Progressive of the page', required=True, readonly=True),
         'progressive_line_number': fields.integer('Progressive line', required=True, readonly=True),
-        'progressive_credit': fields.float('Progressive Credit', digits_compute=dp.get_precision('Account'), required=True, readonly=True),
-        'progressive_debit': fields.float('Progressive Debit', digits_compute=dp.get_precision('Account'), required=True, readonly=True),
+        'progressive_credit': fields.float('Progressive Credit', digits=dp.get_precision('Account'), required=True, readonly=True),
+        'progressive_debit': fields.float('Progressive Debit', digits=dp.get_precision('Account'), required=True, readonly=True),
     }
     
     _defaults = {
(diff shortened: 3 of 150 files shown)


l10n_it_account_tax_kind/model/account_tax_kind.py
```diff
@@ -7,11 +7,21 @@
...
+
+    @api.multi
+    def name_get(self):
+        res = []
+        for item in self:
+            name = item.name
+            if item.display_name:
+                name = item.display_name
+            res.append((item.id, name))
+        return res
```
eLBati: @jackjack82 why both ```display_name``` and ```name_get```?  Isn't ```name_get``` sufficient?

jackjack82:
@eLBati yes, but probably someone referenced it. shall I delete the display_name field with no worries?

eLBati:
```display_name``` is a core magic field , every model has it automatically. Here the module is just overriding it

File: l10n_it_account_tax_kind/model/account_tax_kind.py [hunk 2]
```diff
@@ -7,17 +7,19 @@
...
     _name = 'account.tax.kind'
-    _rec_name = 'display_name'
 
     code = fields.Char(string='Code', size=3, required=True)
     name = fields.Char(string='Name', required=True)
-    display_name = fields.Char(string='Name', compute='_compute_display_name')
 
-    @api.depends('code', 'name')
     @api.multi
-    def _compute_display_name(self):
```
SimoRubi: In order to have `display_name` recomputed when `code` or `name` change, you can add something like
```
@api.depends('code', 'name')
def _compute_display_name(self):
    """ Just add code and name dependencies for display_name recomputation"""
    return super(AccountTaxKind, self)._compute_display_name()
```

l10n_it_account_tax_kind/tests/test_account_tax_kind.py
```diff
@@ -11,12 +11,6 @@ def setUp(self):
         super(TestAccountTaxKind, self).setUp()
         self.tax_kind_n1 = self.env.ref('l10n_it_account_tax_kind.n1')
 
-    def test_compute_display_name(self):
```
SimoRubi: Why removing this?

l10n_it_account_tax_kind/model/account_tax_kind.py
```diff
@@ -7,17 +7,19 @@
...
-        for record in self:
-            record.display_name = u'[%s] %s' % (record.code, record.name)
+    def name_get(self):
+        res = []
+        for item in self:
+            name = item.name
+            if item.display_name:
+                name = item.display_name
+            res.append((item.id, name))
+        return res
```
eLBati: Guardando meglio, questo ```name_get``` sembra non servire a nulla, perchè ora ```display_name``` è sempre uguale a ```name```

l10n_it_account_tax_kind/model/account_tax_kind.py
eLBati: Why would you do this? ```display_name``` is not stored nor viewable in any form

SimoRubi: yes but I see [here](https://github.com/odoo/odoo/blob/11.0/odoo/models.py#L1487) that it depends on the field defined in `_rec_name`, then why not adding the fields we need?

l10n_it_account_tax_kind/model/account_tax_kind.py
SimoRubi: Vero, @jackjack82 secondo me 
```
name = item.name
if item.display_name:
    name = item.display_name
res.append((item.id, name))
```

deve diventare 
`u'[%s] %s' % (item.code, item.name)`
o qualcosa di simile

l10n_it_account_tax_kind/model/account_tax_kind.py
SimoRubi: @jackjack82 we can skip the dependencies as long as the test works

l10n_it_account_tax_kind/model/account_tax_kind.py
SimoRubi: @jackjack82 restore the old structure for the name of the tax: it should be `[code] name`.
Moreover you can simplify everything following [this](https://github.com/odoo/odoo/blob/3656999f5e51658e0611af100c1507abc7647d94/odoo/addons/base/ir/ir_model.py#L711) example

l10n_it_account_tax_kind/tests/test_account_tax_kind.py
SimoRubi: @jackjack82 restore the test, moreover you don't need to call `self.tax_kind_n1._compute_display_name()`, just assert that the `display_name` is what we expect it to be

SimoRubi:
Also @jackjack82, I investigated the warning:
```
WARNING openerp_test py.warnings: /home/travis/odoo-10.0/odoo/models.py:362: UnicodeWarning: Unicode equal comparison failed to convert both arguments to Unicode - interpreting them as being unequal
elif not all(cols[field.name][key] == vals[key] for key in vals):
```

It is raised because of non-ascii characters [here](https://github.com/OCA/l10n-italy/blob/a21adbdc3712f9281e4ba7f9cd6728100b1a30c0/l10n_it_fatturapa/models/company.py#L17) (both `è` and the quotes) and [there](https://github.com/OCA/l10n-italy/blob/0f6e8adf338e8c7714285db91b89fba481d7d6bd/l10n_it_ricevute_bancarie/models/riba_config.py#L21).
In order to remove that warning too, I suggest to use english for the attributes in py files and the 
`it.po` files for the translations to italian.

jackjack82:
@SimoRubi thank you for the hint, it worked!
@eLBati @SimoRubi here all works. I also updated the name_get. 
Merge or close it, no more work on this. thx ;)

SimoRubi:
There still are some open points:

- restore the test on `display_name`
- fix the translations `a'` instead of `à` is not necessary in all the files you edited, moreover it should be `à` in at least the italian translation.

Since you won't work on this anymore, I'll do that in another PR and you can close this.
Thanks anyway for your work!

PR #554, opened 2018-07-17
[10.0][DDT] Add email template and send button in DDT view

robyf70:


l10n_it_ddt/data/mail_template_data.xml
```diff
@@ -0,0 +1,118 @@
...
+<strong>${object.date_done}</strong>
+%endif
+from ${object.company_id.name}.
+</p>
+<p>You can reply to this email if you have any questions.</p>
+<p>Thank you.</p>
+]]></field>
+        </record>
+
+        <record id="mail_template_data_notification_email_ddt" model="mail.template">
```
eLBati: Why 2 email templates?

robyf70: Has been split for convenience, the second template contains header and footer and the first the body

l10n_it_ddt/models/stock_picking_package_preparation.py
```diff
@@ -435,6 +435,44 @@ def action_invoice_create(self):
...
+                                     'email_compose_message_wizard_form')[1]
+        except ValueError:
+            compose_form_id = False
+
+        ctx = {
+            'default_model': 'stock.picking.package.preparation',
+            'default_res_id': self.ids[0],
+            'default_use_template': bool(template_id),
+            'default_template_id': template_id,
+            'mark_so_as_sent': True,
```
SimoRubi: This should be replaced by something like `mark_ddt_as_sent` and used properly, otherwise just remove it

robyf70: right! will have a look

robyf70: Have removed for now the mark_so_as_sent because not used/required in this particular context. I will add it as additional patch on top of this after approved ;)

SimoRubi: Perfect, anyway I still can't manage to see `mail_template_data_notification_email_ddt` working, any advice?

robyf70: If you have received the notification email with attached the DDT, the header and footer are exactly the template we are talking

SimoRubi: I'm trying this in runbot and it doesn't really send emails, is there any way to check that within Odoo (I tried in Email/Email but no luck)

robyf70: Have you tried to run it locally?

SimoRubi: No I haven't, but I've managed to see it rendered in runbot and it looks good.

l10n_it_ddt/data/mail_template_data.xml
```diff
@@ -0,0 +1,118 @@
...
+            <field name="report_template" ref="action_report_ddt"/>
+            <field name="report_name">${(object.ddt_number)}</field>
+            <field name="auto_delete" eval="True"/>
+            <field name="model_id" ref="model_stock_picking_package_preparation"/>
+            <field name="lang">${object.partner_id.lang}</field>
+            <field name="body_html"><![CDATA[
+<p>Dear ${object.partner_id.name},</p>
+<p>Here is your document of
+%if object.date:
+<strong>${object.date},</strong>
```
SimoRubi: To make it perfect, you could try to manage when this `object.date` (respectively `object.date_done`) is not set, because in that case the template would be weird:
```
Here is your document of with shipping date from YourCompany
```

robyf70: Ok! Will have a look later today or tomorrow

robyf70: Right! Wrong branch! Now you have it!

eLBati: It seems to me the problem described by @SimoRubi https://github.com/OCA/l10n-italy/pull/554/files#r204357167 would still be present

robyf70: @eLBati I've double checked it and seems ok to me, no?

SimoRubi: @robyf70 I made https://github.com/LevelPrime/l10n-italy/pull/1 to fix what I meant in https://github.com/OCA/l10n-italy/pull/554#discussion_r204357167

PR #556, opened 2018-07-18
[10.0] IMP l10n_it_fatturapa_out message when fiscal position not set

eLBati:


sbidoul:
Yes, it is better to NOT modify the po files manually. Immediately after merge, the .pot file will be updated by travis, and the .po files will be updated on weblate a couple of hours later.

SimoRubi:
@sbidoul so also editing all the po and the pot might cause issues? 
I've been looking at the [docs](https://docs.google.com/document/d/1b7lzeKyaDJ_zgmZ3VKWNqJyFb5gnsdXF41-DtPKLvRY/edit) of the new weblate process and this part is still not very clear to me:

- When the pot file already exists (this is the case in this PR), weblate pulls it hourly at hh:15 and changes all the po accordingly (it changes po files in Github *on the fly*? How? I can't find where this is done)

Here @eLBati already changed all the po files too so I don't get which issues we might run into.
Thanks

sbidoul:
Weblate 3 has an addon that runs `msgmerge` to update .po from .pot. It then commits the changes to github.

About the conflicts: since weblate can git commit .po files at any time, if someone pushes .po files manually there might be git conflicts. These could be painful to resolve.

So, except for new modules, it should never be necessary to touch .po and .pot files manually. For new modules it is acceptable because there is no risk of conflicts.

Moreover I've seen quite a lot of syntax errors in .po files following manual changes. So it's easier and safer to let weblate do it.

SimoRubi:
Ok thanks now I get why we are not supposed to edit po files: we have a translations platform so it rules over translations.
So in this particular use case (change a string) we as developers are expected to:

1. Change everything we need in the module, except po files
2. After few hours (weblate pulls the pot hourly) that PR is merged, go to weblate and translate it
3. After one day, the translation should be on github

Is this the correct flow @sbidoul?
Thanks!

sbidoul:
Yep, "weblate rules over translations" is a good way to put it :)

eLBati:
@SimoRubi @tafaRU removed changes to po files

PR #583, opened 2018-08-31
[10.0] FIX  l10n_it_codici_carica api.constrains gets a recordset

eLBati:
https://www.odoo.com/documentation/10.0/howtos/backend.html#model-constraints


eLBati:
Per 11.0: https://github.com/OCA/l10n-italy/pull/584

PR #538, opened 2018-06-29
11.0 mig l10n it ddt

SilvioGregorini:

PR #587, opened 2018-09-03
11.0 mig l10n it ddt delivery

SimoRubi:
Depends on https://github.com/OCA/l10n-italy/pull/538

Migration from https://github.com/OCA/l10n-italy/tree/8.0/l10n_it_ddt_delivery (v8)


SimoRubi:
Actually depends on #588

PR #588, opened 2018-09-04
11.0 mig l10n it ddt

SimoRubi:
Superseding https://github.com/OCA/l10n-italy/pull/538

PR #597, opened 2018-09-26
11.0 mig l10n it ddt fix background in report

SilvioGregorini:
@eLBati
This PR is actually "empty", since it has the same commits as the PR I made to @SimoRubi repo, which will add up to #588 
Yet, @dcorio and I noticed that, in v11.0, if the ddt report is printed with 'background' as company's document template, the CSS will make the report spread on two pages. This may happen because of the CSS used for the document, in particular `o_report_layout_background` which sets `min-height: 900px`.
The easiest way to fix this, then, would be to set a t-if conditioned style tag like this:

```
<template id="report_ddt">
    <t t-call="web.external_layout">
        <t t-set="o" t-value="o.with_context({'lang':o.partner_id.lang})" />
        <t t-if="o.company_id.external_report_layout == 'background'">
            <style>
                .o_report_layout_background {
                    min-height: 300px;
                }
            </style>
        </t>
        <div class="page">
```

but I think there must be a better way to do so, maybe even without impacting the default CSS.
Any idea on how to do it?


eLBati:
Essendo il modulo stato portato a 11.0, chiudo

PR #599, opened 2018-09-28
bugfix encoding

luca-vercelli:


luca-vercelli:
I am running odoo 11.0 with python 2. I don't know if that is supported, anyway it works. The only issue I found is this problem on encoding.

SimoRubi:
@luca-vercelli in the [Odoo installation guide](https://www.odoo.com/documentation/11.0/setup/install.html) is stated:

> if Python is already installed, make sure it is 3.5 or above, previous versions are not compatible with Odoo

So Odoo 11 does not support an installation in a Python 2 environment

luca-vercelli:
Ok sorry.

PR #608, opened 2018-10-11
11.0 mig l10n it website sale corrispettivi

SimoRubi:
Standard migration


tafaRU:
@SimoRubi, I see WIP in the PR name. Is that still true?

SimoRubi:
@tafaRU removed the WIP 'status', thanks

l10n_it_website_sale_corrispettivi/tests/test_website_sale_no_corrispettivi.py
```diff
@@ -0,0 +1,30 @@
...
+
+    def test_admin_no_corrispettivi(self):
+        """
+        If a partner having the flag 'corrispettivi' set to False
+        creates an order in frontend, then
+        the created order's flag corrispettivi is False"""
+        self.assertFalse(self.admin_partner.use_corrispettivi)
+        existing_orders = self.env['sale.order'].search([])
+        # In frontend, create an order with flag corrispettivi
+        self.phantom_js(
```
tafaRU: Curiosity: wouldn't be possible call directly https://github.com/odoo/odoo/blob/11.0/addons/website_sale/tests/test_sale_process.py#L9 ?

SimoRubi: @tafaRU I think that would be possible but that might be not a good idea, another test (in this case https://github.com/odoo/odoo/blob/11.0/addons/website_sale/tests/test_sale_process.py#L9) might have its own setup that does not get called if I simply write `test_sale_process()`

tafaRU: Ok, thanks for the explanation.

SimoRubi:
@tafaRU you are right it doesn't work with the public user and it never has.
I have edited the README in https://github.com/OCA/l10n-italy/pull/608/commits/91fdd45c6f58e758257e96425f428c8e298c5e03 and added a few fixes considering that the partner might be the public user

l10n_it_website_sale_corrispettivi/controllers/main.py
```diff
@@ -8,21 +7,22 @@
...
-        res = super(WebsiteSaleCorrispettivi, self) \
+        partner_id = super(WebsiteSaleCorrispettivi, self) \
             ._checkout_form_save(mode, checkout, all_values)
         use_invoice = all_values.get('use_invoice')
         # Update order
         order = request.website.sale_get_order()
-        order_values = \
-            self._prepare_corrispettivi_order_values(order, use_invoice)
-        order.write(order_values)
+        order_values = self.sudo()._prepare_corrispettivi_order_values(
```
eLBati: This raises
```
l10n_it_website_sale_corrispettivi/controllers/main.py", line 15, in _checkout_form_save
    order_values = self.sudo()._prepare_corrispettivi_order_values(
AttributeError: 'WebsiteSale (extended by WebsiteSaleCorrispettivi)' object has no attribute 'sudo'
```

Also, using sudo here, in a multi-company environment, if `order.company_id` was empty, this line https://github.com/OCA/l10n-italy/blob/3cc6525fca3a9f1a95c831acaf3e0d208b40bcb5/l10n_it_website_sale_corrispettivi/controllers/main.py#L54 would lead to use the company of admin user, running into undesired behaviour potentially

SimoRubi: Thanks for your review!
I removed the `sudo()` call for `self` in https://github.com/OCA/l10n-italy/pull/608/commits/79bf9cbee01a1cfa863077c6ec4a870e65163791.
I also changed the *company behavior*: I will always look for the fiscal position having the company of the order (or partner), even if it is `False`.

eLBati: Now I get
```
  File "/home/elbati/workspace/odoo/instances/odoo11/parts/l10n-italy/l10n_it_website_sale_corrispettivi/controllers/main.py", line 16, in _checkout_form_save
    order, use_invoice)
  File "/home/elbati/workspace/odoo/instances/odoo11/parts/l10n-italy/l10n_it_website_sale_corrispettivi/controllers/main.py", line 55, in _prepare_corrispettivi_order_values
    .get_corr_fiscal_pos(company).id
  File "/home/elbati/workspace/odoo/instances/odoo11/parts/l10n-italy/l10n_it_corrispettivi/models/account.py", line 94, in get_corr_fiscal_pos
    ('company_id', '=', company_id.id)], limit=1)
  File "/home/elbati/workspace/odoo/instances/odoo11/parts/odoo/odoo/models.py", line 1480, in search
    res = self._search(args, offset=offset, limit=limit, order=order, count=count)
  File "/home/elbati/workspace/odoo/instances/odoo11/parts/odoo/odoo/models.py", line 3767, in _search
    self.sudo(access_rights_uid or self._uid).check_access_rights('read')
  File "/home/elbati/workspace/odoo/instances/odoo11/parts/odoo/odoo/models.py", line 2862, in check_access_rights
    return self.env['ir.model.access'].check(self._name, operation, raise_exception)
  File "<decorator-gen-20>", line 2, in check
  File "/home/elbati/workspace/odoo/instances/odoo11/parts/odoo/odoo/tools/cache.py", line 89, in lookup
    value = d[key] = self.method(*args, **kwargs)
  File "/home/elbati/workspace/odoo/instances/odoo11/parts/odoo/odoo/addons/base/ir/ir_model.py", line 1201, in check
    raise AccessError(msg % msg_params)
odoo.exceptions.AccessError: ('Sorry, you are not allowed to access this document. Only users with the following access level are currently allowed to do that:\n- Accounting & Finance/Billing Manager\n\t- Employees/Employee\n\t- Other Extra Rights/Portal\n\n(Document model: account.fiscal.position)', None)
```

SimoRubi: I tested that in my local environment and fixed that with https://github.com/OCA/l10n-italy/pull/608/commits/5d9bc4087adcad8c77865c5fea915091133a06d8.
The issue was that I was confusing the current user (public user) and the partner that had been created in the checkout: so the public user was not able to read/edit the partner that had just been created.
Anyway, I specified that in the comments.

Again, thanks for the review! :)

SimoRubi:
Fixed up

PR #628, opened 2018-10-31
[12.0] mig l10n_it_rea

eLBati:
and add tests


SimoRubi:
@tafaRU having 2 positive reviews, I think this can be merged

Issue #654, opened 2018-11-09
[10.0] l10n_it_ddt: record does not exist or has been deleted


SimoRubi:
Steps:

1. Create a sale order (sell a Computer case to Agrolait) and check _Automatically create a DDT_
2. Confirm the order
3. Open the DDT
4. Open the transfer inside the DDT and Force Availability > Validate
5. Go back to the DDT and try to set _Reason for transportation_

Actual behavior:
Error: **Record does not exist or has been deleted.**

Desired behavior:
no error

PR #656, opened 2018-11-09
[10.0][FIX] l10n_it_ddt: Fix https://github.com/OCA/l10n-italy/issues/654

SimoRubi:
Fix https://github.com/OCA/l10n-italy/issues/654

### DIFF
--- a/l10n_it_ddt/models/stock.py
+++ b/l10n_it_ddt/models/stock.py
@@ -61,3 +61,15 @@ class StockPicking(models.Model):
                 return self.partner_id
         else:
             return self.location_dest_id.partner_id
+
+    @api.multi
+    def open_form_current(self):
+        self.ensure_one()
+        return {
+            'type': 'ir.actions.act_window',
+            'view_type': 'form',
+            'view_mode': 'form',
+            'res_model': self._name,
+            'res_id': self.id,
+            'target': 'current'
+        }
--- a/l10n_it_ddt/models/stock.py
+++ b/l10n_it_ddt/models/stock.py
@@ -5,8 +5,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 
-from odoo import fields, models, api, _
-from odoo.exceptions import ValidationError
+from odoo import fields, models, api
 
 
 class StockPicking(models.Model):
@@ -20,22 +19,6 @@ class StockPicking(models.Model):
         column2='stock_picking_package_preparation_id',
         string='DdT',
         copy=False, )
-
-    @api.multi
-    @api.constrains('state')
-    def _ddt_constrain_state(self):
-        """Prevent pickings that are included in a DDT to be set as done
-        unless it is being processed by DDT"""
-        if self.env.context.get('ddt_processing', False):
-            return
-        for picking in self:
-            if picking.state == 'done' and picking.ddt_ids:
-                raise ValidationError(
-                    _('Cannot set to "Done" picking %s '
-                      'because it is included in DDT %s.\n'
-                      'Process the DDT or unlink the picking from the DDT.')
-                    % (picking.display_name,
-                       picking.ddt_ids[0].display_name))
 
     @api.multi
     def write(self, values):
--- a/l10n_it_ddt/models/stock.py
+++ b/l10n_it_ddt/models/stock.py
@@ -5,7 +5,8 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 
-from odoo import fields, models, api
+from odoo import fields, models, api, _
+from odoo.exceptions import ValidationError
 
 
 class StockPicking(models.Model):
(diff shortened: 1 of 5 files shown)


l10n_it_ddt/models/stock.py
```diff
@@ -20,6 +21,22 @@ class StockPicking(models.Model):
...
+        """Prevent pickings that are included in a DDT to be set as done
+        unless it is being processed by DDT"""
+        if self.env.context.get('ddt_processing', False):
+            return
+        for picking in self:
+            if picking.state == 'done' and picking.ddt_ids:
+                raise ValidationError(
+                    _('Cannot set to "Done" picking %s '
+                      'because it is included in DDT %s.\n'
+                      'Process the DDT or unlink the picking from the DDT.')
```
eLBati: @SimoRubi questo non è un caso d'uso da escludere dal modulo. L'utente deve poter processare i picking indipendentemente dal DDT

SimoRubi: Fatto in https://github.com/OCA/l10n-italy/pull/656/commits/d4f7bd8ce7b584d6e387e4385d86b0d55bf5a946

SimoRubi:
@eLBati ho eliminato il vincolo sul picking, in effetti per impedire il messaggio di errore è sufficiente utilizzare il widget `one2many_list`

joezsweet:
Tested, forcing available qty on picking. Now rows are not copied in DDT details:

schermata 2018-11-20 alle 09 48 26

schermata 2018-11-20 alle 09 48 34

Related Picking

schermata 2018-11-20 alle 09 50 58

SimoRubi:
@joezsweet I can't replicate your issue, could you provide the steps or a video?
Here what I did:

1. Create a sale order (sell an Ice cream to Agrolait) and check Automatically create a DDT
2. Confirm the order
3. Open the DDT
4. Open the transfer inside the DDT and Force Availability
5. It warns that some fields are missing in the DDT, so edit, fill them and save
6. Open the transfer inside the DDT and Force Availability > Validate

After step 6 I still see the details of the DDT

joezsweet:
Have a look  at: http://recordit.co/Tbt6s4bKcv

SimoRubi:
Thanks for the video @joezsweet, now I managed to replicate your issue.
I discovered that it is actually an issue of one of the base modules for l10n_it_ddt, I created https://github.com/OCA/stock-logistics-workflow/pull/472 to fix it.

joezsweet:
@SimoRubi it seems 
https://github.com/OCA/stock-logistics-workflow/pull/472 has been merged, next step?

l10n_it_ddt/views/stock_picking_package_preparation.xml
```diff
@@ -85,6 +85,9 @@
                        domain="['|', ('id',  '=', partner_id), ('parent_id', '=', partner_id)]"
                        attrs="{'invisible':[('ddt_type_id', '=', False)], 'required':[('ddt_type_id', '!=', False)]}"/>
             </field>
+            <field name="picking_ids" position="attributes">
+                <attribute name="widget">one2many_list</attribute>
+            </field>
```
eLBati: @SimoRubi this prevents the user to add existing available pickings in `Transfers` tab of DDT

eLBati:
@SimoRubi is the original error raised by `_compute_to_be_invoiced` ? Did you debug that?

SimoRubi:
Rebased on `10.0`

l10n_it_ddt/views/stock_picking_package_preparation.xml
SimoRubi: @eLBati verified, then the widget change cannot be a solution

SimoRubi:
@eLBati the issue has to do with the consistency of `stock.picking.package.preparation` in the form view.
The consistency is compromised as soon as a `stock.move` in the `stock.picking` is edited because all the lines `stock.picking.package.preparation.line` are deleted and recreated by `_update_line_ids` in the overridden `stock.move.write`, but the form is keeping the old instances of `stock.picking.package.preparation.line`.

eLBati:
@SimoRubi so it should be related to https://github.com/odoo/odoo/issues/17618

Which is the involved onchange?

In `sale_commission` module, where we got into https://github.com/odoo/odoo/issues/17618, we replaced the onchange with a button that recomputed the lines

SimoRubi:
@eLBati as soon as the picking changes its state, all the lines in the tab _details_ are substituted, but in the form the old ones remain; this happens in the base module too but it does not trigger any error because it has nothing to compute.

In _l10n_it_ddt_ instead the error is raised when the field `sale_line_id` is (trying to be) recomputed, but I am afraid that adding any computed field to the line would raise the same error, so I don't think that adding a **Recompute** button can fix this.

I am now looking for a way to:

1. Prevent opening the picking as a popup (the _no_open_ option apparently only works in m2o fields);
2. Adding a button to edit the picking in a new form (like when you select an element in the listview), but that should be covered by https://www.odoo.com/forum/help-1/question/how-to-open-one2many-form-view-of-the-correspondent-line-in-tree-view-112388

SimoRubi:
@joezsweet could you please check if the recent modifications fix the issue? Thanks!

SimoRubi:
@eLBati can you update your review? Thanks

joezsweet:
@SimoRubi it seems to be working fine.  Any chance to use wizard to show picking from ddt?

dcorio:
@joezsweet what do you mean with "to show picking from ddt"? pickings are listed on a dedicated tab on the DdT form.

SimoRubi:
@joezsweet could you please add a positive review if you checked that it works?
I'm afraid that in the current status of Odoo core, the popup can't be restored safely.

The technical issue here is that the change made inside the popup changes also the data that the form refers to (in particular the _details' lines_), but the form _doesn't notice_.
It is a known issue of Odoo core (https://github.com/odoo/odoo/issues/17618 as stated by @eLBati).

The most simple workaround that I found is to open the picking in the current page instead of the popup.

joezsweet:
@SimoRubi just le me know how to add a positive review 😬 

Did u check:
https://github.com/OCA/web/tree/11.0/web_ir_actions_act_view_reload

Could that do the job?

SimoRubi:
@joezsweet thanks I did not know that module but in our case it doesn't work because we are not calling the backend, thus we can't call the action defined in that module.

In order to review, see https://github.com/OCA/l10n-italy/pull/588#issuecomment-437322910 and just click on _Approve_ in the 4th step instead of _Request changes_.

PR #682, opened 2018-11-28
l10n_it_fatturapa_pec: check possible fetch errors and track them

eLBati:
While sending, check for valid fetchmail server


l10n_it_fatturapa_pec/models/fatturapa_attachment_out.py
```diff
@@ -47,8 +47,19 @@ def reset_to_ready(self):
                 raise UserError(_("Yo can only reset 'sender error' files"))
```
SimoRubi: _Yo_ -> _You_

eLBati: done

PR #684, opened 2018-11-29
FIX l10n_it_fatturapa: partner checks must be done for customers only…

eLBati:
…, as only needed for XML generation


eLBati:
Però aspettiamo questa https://github.com/OCA/l10n-italy/pull/677

PR #710, opened 2018-12-10
[12.0] fix electronic invoice translations

eLBati:


sbidoul:
Ok, I have pushed all translations from weblate, and I merge this. There should be no conflict, let's see.

sbidoul:
All went well, it's looking good. 229 strings for l10n_it_fatturapa in weblates, so it caught up now.

PR #748, opened 2018-12-19
[10.0][FIX] attachment name max lenght 60

sergiocorato:
v. 8.0 #745


l10n_it_fatturapa_out/wizard/wizard_export_fatturapa.py
```diff
@@ -721,8 +721,12 @@ def setDatiPagamento(self, invoice, body):
     def setAttachments(self, invoice, body):
         if invoice.fatturapa_doc_attachments:
             for doc_id in invoice.fatturapa_doc_attachments:
+                nome_attachment = doc_id.datas_fname if \
+                   len(doc_id.datas_fname) <= 60 else \
+                   ''.join([doc_id.name.rsplit('.', 1)[0][:55],
+                            doc_id.name.rsplit('.', 1)[1]])
```
eLBati: I would use `os.path.splitext` to get the file name without extension

sergiocorato: ok, I add os import

l10n_it_fatturapa_out/wizard/wizard_export_fatturapa.py
```diff
@@ -721,8 +722,12 @@ def setDatiPagamento(self, invoice, body):
     def setAttachments(self, invoice, body):
         if invoice.fatturapa_doc_attachments:
             for doc_id in invoice.fatturapa_doc_attachments:
+                nome_attachment = doc_id.datas_fname if len(
+                    doc_id.datas_fname) <= 60 else os.path.join([
```
eLBati: Why `os.path.join` here? I don't think it gives the desired result. You can use `'.'.join`

l10n_it_fatturapa_out/wizard/wizard_export_fatturapa.py
```diff
@@ -721,8 +722,12 @@ def setDatiPagamento(self, invoice, body):
...
+                        os.path.splitext(doc_id.name)[0][:55],
```
eLBati: Assuming there will not be extensions longer than 5 chars?

sergiocorato: is it possible? I change the function

l10n_it_fatturapa_out/wizard/wizard_export_fatturapa.py
```diff
@@ -721,8 +722,14 @@ def setDatiPagamento(self, invoice, body):
     def setAttachments(self, invoice, body):
         if invoice.fatturapa_doc_attachments:
             for doc_id in invoice.fatturapa_doc_attachments:
+                nome_attachment = doc_id.datas_fname if len(
+                    doc_id.datas_fname) <= 60 else '.'.join([
```
eLBati: But
```
>>> import os
>>> name = 'test.zip'
>>> '.'.join([os.path.splitext(name)[0][:(60-len(os.path.splitext(name)[1]))],os.path.splitext(name)[1]])
'test..zip'
```

so better `''.join`

sergiocorato: fatto

sergiocorato:
fatto

eLBati:
@sergiocorato riesci a farlo anche per v11 e v12?

PR #751, opened 2018-12-19
[11.0] ADD TD20 doc type - set default document type in invoice

labaggio:
IMP l10n_it_fiscal_document_type to be sure that doc type is always set


labaggio:
@eLBati non mergiamo? :)

PR #796, opened 2019-01-08
[12.0] FIX l10n_it_fatturapa when receiving e-invoices signed and base64 encrypted

eLBati:


eLBati:
Chiudo in favore della https://github.com/OCA/l10n-italy/pull/899

PR #815, opened 2019-01-09
[12.0] FIX l10n_it_fatturapa allowing to import XML elements with blank spaces, as already done for String*LatinType

eLBati:
Strings of blank spaces are allowed by SDI.
{1,60} would produce
SimpleFacetValueError: Type
{http://ivaservizi.agenziaentrate.gov.it/docs/xsd/fatture/v1.2}
String60Type pattern constraint violated by value


primes2h:
Test ok :+1:

PR #817, opened 2019-01-10
[10.0] FIX l10n_it_account_stamp: recompute taxes when adding stamp to lines

eLBati:
This is needed as stamp line should be under VAT exemption

v12: https://github.com/OCA/l10n-italy/pull/818

### DIFF
--- a/l10n_it_account_stamp/__manifest__.py
+++ b/l10n_it_account_stamp/__manifest__.py
@@ -7,7 +7,7 @@
 
 {
     'name': 'Italian Localization - Tax Stamp',
-    'version': '10.0.1.0.0',
+    'version': '10.0.1.0.1',
     'category': 'Localization/Italy',
     'summary': 'Tax stamp automatic management',
     'author': 'Ermanno Gnan, Sergio Corato, Enrico Ganzaroli, '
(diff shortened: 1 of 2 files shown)


eLBati:
> Could you please add a test?

@SimoRubi The module does not have tests at all and at the moment I don't have time to add them.

> why not take advantage of https://github.com/odoo/odoo/blob/10.0/addons/account/models/account_invoice.py#L457-L463 (that is what happens when you manually add an invoice line)?

Because `compute_taxes` has been created for this scope and calling an `onchange` method would not write to DB.

SimoRubi:
A little more research makes me realize that you are right and `compute_taxes` is generally used when adding a new invoice line.
My concern was that in the method it is specified that it is to be used on a **fresh** (just created) invoice. Too bad it is not done automatically by Odoo core (at least [they know that too](https://github.com/odoo/odoo/blob/10.0/addons/sale/models/sale.py#L386-L387)).

Too bad for automated tests too, ok then.

tafaRU:
@SimoRubi please also review https://github.com/OCA/l10n-italy/pull/818

SimoRubi:
Does this need to be backported to v8 too?

PR #824, opened 2019-01-10
[8.0][IMP] update fatturapa_out version

SimoRubi:
This https://github.com/OCA/l10n-italy/commit/57c8bfdeb39106b2c90f5b0b5d44ab2847db4b07#diff-38398b63046830d1d3dec4b6921cd662 caused some pypi builds with version 10.0.2.0.0 that are actually older than the current version.


SimoRubi:
This fixes #822 (where the version 2.0.0 was installed)

PR #846, opened 2019-01-14
[12.0] REF l10n_it_fatturapa_in for cases of partners (contacts) with wrong fiscal code (=TIN)

eLBati:
First search by VAT number, otherwise by fiscal code.
Correctly check commercial partner

### DIFF
--- a/l10n_it_fatturapa_in/wizard/wizard_import_fatturapa.py
+++ b/l10n_it_fatturapa_in/wizard/wizard_import_fatturapa.py
@@ -116,26 +116,32 @@ class WizardImportFatturapa(models.TransientModel):
                 DatiAnagrafici.IdFiscaleIVA.IdPaese,
                 DatiAnagrafici.IdFiscaleIVA.IdCodice
             )
-        partners = partner_model.search([
-            '|',
-            ('vat', '=', vat or 0),
-            ('fiscalcode', '=', cf or 0),
-        ])
-        commercial_partner = False
+        partners = partner_model
+        if vat:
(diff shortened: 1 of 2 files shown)


sergiocorato:
consiglio queste modifiche 8c8dad364e67ff26fcf388f45badd9d6b91489db

eLBati:
@sergiocorato @monen17 grazie, fatto.

l10n_it_fatturapa_in/wizard/wizard_import_fatturapa.py
```diff
@@ -116,26 +116,32 @@ def getPartnerBase(self, DatiAnagrafici):
                 DatiAnagrafici.IdFiscaleIVA.IdPaese,
                 DatiAnagrafici.IdFiscaleIVA.IdCodice
             )
-        partners = partner_model.search([
-            '|',
-            ('vat', '=', vat or 0),
-            ('fiscalcode', '=', cf or 0),
-        ])
-        commercial_partner = False
+        partners = partner_model
```
GSLabIt: Can I ask why this? would't be better partners= False? Am I missing something?

eLBati: `partner_model` is an empty set. Using `partners= False` , line `len(partners)` would raise exception

GSLabIt: Agreed, in that case would be possible to check `if partners and len(partners) > 1`. Btw it's ok, thx

PR #852, opened 2019-01-16
[10.0][IMP][l10n_it_fatturapa_out] Do not export descriptive invoice lines

robyf70:
Account invoice permit to have descriptive lines with quantity = 0, price_unit = 0 and no taxes specified.
This patch filter out such lines and avoid ADE to complain and report errors. 

As side effect this patch add an hook to permit module to extend or change the filtering logic.

### DIFF
--- a/l10n_it_fatturapa_out/wizard/wizard_export_fatturapa.py
+++ b/l10n_it_fatturapa_out/wizard/wizard_export_fatturapa.py
@@ -592,7 +592,7 @@ class WizardExportFatturapa(models.TransientModel):
             'Product Unit of Measure')
         if uom_precision < 2:
             uom_precision = 2
-        for line in invoice.invoice_line_ids:
+        for line in self.getAccountInvoiceLines(invoice):
             if not line.invoice_line_tax_ids:
                 raise UserError(
                     _("Invoice line %s does not have tax.") % line.name)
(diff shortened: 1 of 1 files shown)


sergiocorato:
Hi @robyf70 invoice is correctly validated from SdI even with quantity 0, price_unit 0, only taxes are needed. More, in this way you don't send some lines to your customer, I think is not you intended to do.

robyf70:
@sergiocorato Right! Anyway, the problem is, if we have descriptive lines without tax, which is possible with account.invoice and we have customers doing that for describing services, filtering them out is acceptable for ADE and will be the same for the customer as soon as a PDF copy of the invoice is attached with the PEC, no?

sergiocorato:
> @sergiocorato Right! Anyway, the problem is, if we have descriptive lines without tax, which is possible with account.invoice and we have customers doing that for describing services, filtering them out is acceptable for ADE and will be the same for the customer as soon as a PDF copy of the invoice is attached with the PEC, no?

I think it's not the best way, as it presumes that a customer opens the pdf and manually writes some datas in the invoice, losing time or making some mistakes. The best way i think is to force customer to always put tax in invoice lines, as it can be done with [this](https://github.com/OCA/account-financial-tools/tree/10.0/account_invoice_tax_required) module

robyf70:
@sergiocorato Agree! But don't forget we are talking about description lines which doesn't impact at all the accounting. Setting tax field for each description mean to have a "fuori campo iva" tax code which will "clutter" every description line. Such lines aren't formally "fuori campo iva", so they are wrongly "classified" if we declare them as such. That's my interpretation, but indeed I can be wrong :)

sergiocorato:
> @sergiocorato Agree! But don't forget we are talking about description lines which doesn't impact at all the accounting. Setting tax field for each description mean to have a "fuori campo iva" tax code which will "clutter" every description line. Such lines aren't formally "fuori campo iva", so they are wrongly "classified" if we declare them as such. That's my interpretation, but indeed I can be wrong :)

Why 'fuori campo iva'? :D this line can be default IVA, they are 0 so IVA is 0 so it doesn't impact in anyway. Removing informations is deprecable if not needed

PR #900, opened 2019-01-25
[12.0] FIX l10n_it_fatturapa: disable mindate check

eLBati:
pyxb is unable to check oder of dates if only one has timezone
the check for date > 1970-01-01 is disabled as a workaround

also modified test xml

TODO
~~https://github.com/OCA/l10n-italy/pull/892#pullrequestreview-196977085~~


eLBati:
Chiudo in favore della https://github.com/OCA/l10n-italy/pull/1050

PR #909, opened 2019-01-29
[10.0] FIX import XML with invalid URI

eLBati:
Recovering parser is needed for files where strings like
`xmlns:ds="http://www.w3.org/2000/09/xmldsig#&quot;"`
are present: even if lxml raises
`{XMLSyntaxError}xmlns:ds: 'http://www.w3.org/2000/09/xmldsig#"' is not a valid URI`
such files are accepted by SDI

Fixes https://github.com/OCA/l10n-italy/issues/905
Fixes https://github.com/OCA/l10n-italy/issues/902

v11: https://github.com/OCA/l10n-italy/pull/910
v12: https://github.com/OCA/l10n-italy/pull/911

### DIFF
--- a/l10n_it_fatturapa_in/__manifest__.py
+++ b/l10n_it_fatturapa_in/__manifest__.py
@@ -7,7 +7,7 @@
 
 {
     'name': 'Italian Localization - Fattura elettronica - Ricezione',
-    'version': '10.0.1.4.1',
+    'version': '10.0.1.4.2',
     'category': 'Localization/Italy',
     'summary': 'Ricezione fatture elettroniche',
     'author': 'Agile Business Group, Innoviu, '
--- a/l10n_it_fatturapa/__manifest__.py
+++ b/l10n_it_fatturapa/__manifest__.py
@@ -6,7 +6,7 @@
 
 {
     'name': 'Italian Localization - Fattura elettronica - Base',
-    'version': '10.0.2.3.2',
+    'version': '10.0.2.3.3',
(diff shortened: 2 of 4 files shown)


eLBati:
@SimoRubi done

l10n_it_fatturapa_in/__manifest__.py
```diff
@@ -7,7 +7,7 @@
 
 {
     'name': 'Italian Localization - Fattura elettronica - Ricezione',
-    'version': '10.0.1.4.1',
+    'version': '10.0.1.4.2',
```
SimoRubi: In my opinion, it is not needed to increment version when adding a test.

tafaRU: I share this view.

eLBati: So you want to produce a `dev` release?

SimoRubi: yes @eLBati, in my opinion adding/editing a test does not deserve a new version: new versions are for new behaviors or bug fixes.

eLBati: @taifu @Byloth @labaggio what do you think?

labaggio: In this case, I can agree..  only this case ;)

taifu: 😂 😂

PR #918, opened 2019-01-30
[10.0] Fix einvoice extraction when the p7m file is already base64 encoded

eLBati:
From https://github.com/OCA/l10n-italy/pull/899


SimoRubi:
@eLBati penso anche io si possa chiudere in favore di

PR #989, opened 2019-02-10
[12.0] FIX If there are characters '\t' (tab) on description of invoice line, the invoice can't be exported

eLBati:
From https://github.com/OCA/l10n-italy/pull/781


eLBati:
Chiudo in favore di https://github.com/OCA/l10n-italy/pull/990

PR #1012, opened 2019-02-13
[8.0] [FIX] workaround for pyxb bug in mindate check

sherpya:
pyxb is unable to check order of dates if only one has timezone
checking if the date > 1970-01-01
we remove tzinfo from parsed date only types as a workaround


sergiocorato:
Chiudo in quanto già fatto upgrade a e-invoice 1.6

PR #1027, opened 2019-02-19
[11.0] [FIX] l10n_it_corrispettivi_sale: use_corrispettive not correcly propagated from partner

GSLabIt:


GSLabIt:
@monen17 try:

1. set use corrispettivo in partner
2. create a sale order for the partner
3. from sale order create a receipt with down payment.
4. check journal on receipt (actually it create an invoice instead of recepit).

eLBati:
Visto che https://github.com/OCA/l10n-italy/issues/1045 è chiusa, chiudo anche questa

PR #1031, opened 2019-02-21
[10.0][FIX] withholding amount - support customer/supplier invoices

sergiocorato:


sergiocorato:
superseed #482 and #676

PR #1035, opened 2019-02-22
10.0 IMP Add xml supplier in search xml

jado95:
v8: https://github.com/OCA/l10n-italy/pull/1034


eLBati:
@jado95 :+1: 
Puoi incrementare la versione?

PR #1078, opened 2019-03-06
[10.0][IMP][l10n_it_fiscalcode] Add fiscalcode in partner search view

robyf70:
This PR simply add `fiscalcode` to partner search view.

### DIFF
--- a/l10n_it_fiscalcode/view/fiscalcode_view.xml
+++ b/l10n_it_fiscalcode/view/fiscalcode_view.xml
@@ -14,4 +14,14 @@
             </xpath>
         </field>
     </record>
+    <record id="view_res_partner_filter_fiscalcode_data" model="ir.ui.view">
+        <field name="name">res.partner.fiscalcode.filter</field>
+        <field name="model">res.partner</field>
+        <field name="inherit_id" ref="base.view_res_partner_filter"/>
+        <field name="arch" type="xml">
+            <search>
+                <field name="fiscalcode"/>
+            </search>
+        </field>
+    </record>
 </odoo>
--- a/l10n_it_fiscalcode/__manifest__.py
+++ b/l10n_it_fiscalcode/__manifest__.py
@@ -5,7 +5,7 @@
 
 {
     'name': 'Italian Localisation - Fiscal Code',
-    'version': '10.0.1.2.0',
+    'version': '10.0.1.3.0',
     'category': 'Localisation/Italy',
     'author': "Odoo Italia Network, Odoo Community Association (OCA)",
     'website': 'https://odoo-community.org/',


l10n_it_fiscalcode/view/fiscalcode_view.xml
```diff
@@ -14,4 +14,14 @@
             </xpath>
         </field>
     </record>
+    <record id="view_res_partner_filter_fiscalcode_data" model="ir.ui.view">
+        <field name="name">res.partner.fiscalcode.filter</field>
+        <field name="model">res.partner</field>
+        <field name="inherit_id" ref="base.view_res_partner_filter"/>
+        <field name="arch" type="xml">
+            <search>
```
tafaRU: why didn't you use `filter` instead of `search`?

robyf70: Do you mean replace `search` with `filter`?

tafaRU: yes!

robyf70: and ... because we have only one entry?

tafaRU: You're inheriting so there is no need to use `<search>`,  tag `<filter>` is enough.

robyf70: I'm using tag `<field>` to search over `<fiscalcode>`, but doesn't matter for what we are looking at. However doesn't work, maybe I'm missing something? See below :

ParseError: Errore durante la validazione del vincolo

L'elemento `<field name="fiscalcode">` non pu\xf2 esser localizzato nella vista genitore

Contesto errore:
Vista `res.partner.fiscalcode.filter`
[view_id: 3073, xml_id: l10n_it_fiscalcode.view_res_partner_filter_fiscalcode_data, model: res.partner, parent_id: 125]
None" while parsing /OtherData/Projects/oca/l10n-italy/l10n_it_fiscalcode/view/fiscalcode_view.xml:17, near
<record id="view_res_partner_filter_fiscalcode_data" model="ir.ui.view">
        <field name="name">res.partner.fiscalcode.filter</field>
        <field name="model">res.partner</field>
        <field name="inherit_id" ref="base.view_res_partner_filter"/>
        <field name="arch" type="xml">
            <field name="fiscalcode"/>
        </field>
    </record>

tafaRU: I meant something like:

```
    <record id="view_res_partner_filter_fiscalcode_data" model="ir.ui.view">
        <field name="name">res.partner.fiscalcode.filter</field>
        <field name="model">res.partner</field>
        <field name="inherit_id" ref="base.view_res_partner_filter"/>
        <field name="arch" type="xml">
            <filter name="parent_id" position="after">
                <field name="fiscalcode"/>
            </filter>
        </field>
    </record>
```

But your solution works just the same so :+1:

eLBati: How can this work without `position`?

robyf70: Guys, what do you think if we can also add the `vat` field as searchable? This is one of the most often requested field to be searchable in partner and I haven't really found the right place, on `l10n_it_*` modules, where to add it.

tafaRU: I tried locally: it  works and all filters seem to be kept.
I also found a similar case: https://github.com/odoo/odoo/blob/10.0/addons/website_sale/views/backend.xml#L29-L31

SimoRubi: It works because the default for `position` is `inside`: https://github.com/odoo/odoo/blob/a409627602ab6c03364b492b5ba7a6fbaba6c5d6/odoo/addons/base/ir/ir_ui_view.py#L554

robyf70:
It works because default `position="inside"` like @SimoRubi said, but if you want to make it explicit we will go for the standard way that @tafaRU brought up. Just let me know what you prefer.

eLBati:
Thanks, didn't know

PR #1325, opened 2019-06-30
[12.0] FIX l10n_it_ddt: quantity_by_lot to correctly compute quantities

eLBati:
See https://github.com/OCA/l10n-italy/issues/1326

Case fixed (see test case):

 - create product tracked by lot
 - set its inventory with qty 10 for lot1 and 10 for lot2
 - create a picking with quantity 6
 - deliver 3 products of lot1 and 3 of lot2
 - create another picking, with quantity 6
 - deliver 4 products of lot1 and 2 of lot2
 - create DDT from the last picking

Print PDF DDT and see wrong quantities: 7 and 5 instead of 4 and 2

--


eLBati:
@SimoRubi fatto

PR #1569, opened 2019-12-22
[12.0]Negative quantity on electonic invoice blocked by SDI (Notifica di scarto)

BitStab:
Descrizione del problema o della funzionalità:
If you have a negative quantity on the invoice to correct the stock with total of 0 the file gets blocked by SDI. This behaviour isn't useful if you f.e. use customizable products using modifiers.

Comportamento attuale prima di questa PR:
Notifica di scarto for electronic invoices with quantities < 0

Comportamento desiderato dopo questa PR:
Negative quantities get replaced with 0 quantity and 0 total to make the xml files pass the tests applied by SDI.

--


eLBati:
This would entail the invoice total to be wrong, doesn't it?

BitStab:
Not quite. The case I'm running into is the following:
Restaurant Orders with some extra wishes of the customers. Those wishes translate into modifiers applied to the specified products, which modify the regular price of the product, but still get registered to have the items correctly moved in the warehouse. The modifiers themselves have no price, so they don't affect the total of an invoice. 
The possible wishes/modifiers can be added or removed from the original product. Removing translates into a new recipe without the selected component. So it also changes the pickings on stock. On the invoice then I have a negative quantity of component X with price 0. It could be done as description, but in this case i would lose the corrisponding stock movements.

Given the fact, that's not possible to send negative quantity lines to SDI in an invoice and given the fact, a negative total on a sales line should never exist, apart from discounts, it's quite straight forward to replace them with quantity 0 on the SDI xml to avoid those error-messages.

PR #1697, opened 2020-03-20
[10.0] install dependencies with pip

sbidoul:
Attempt to workaround sqlalchemy installation issue, by not installing it.

--

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ addons:
 
 env:
   global:
-  - VERSION="10.0" TESTS="0" LINT_CHECK="0" MAKEPOT="0"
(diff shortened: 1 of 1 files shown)


sbidoul:
@eLBati @SimoRubi this is looking good. This does not solve the root cause with sqlalchemy but at least it alleviates the problem for projects that do not need it. FYI, this is how it works:

https://github.com/OCA/maintainer-quality-tools/blob/6ab7d44dc278717fbf7a1f080f5a29de631ef98d/travis/travis_install_nightly#L121-L129

sbidoul:
> I didn't know the whole method could be avoided.

It's very new :) 

> Do you happen to know what is the root cause?

Nope, unfortunately.

eLBati:
@sbidoul thanks!

Issue #1700, opened 2020-03-22
[10.0] move python dependencies out of .travis.yml


sbidoul:
Hi,

I noticed this project installs [python dependencies in `.travis.yml`](https://github.com/OCA/l10n-italy/blob/7bc03a13543eb0e77bd38d7d0bd30882ba8dda8b/.travis.yml#L28-L32).

I'd encourage you to move them to addon manifests.

When that is done, published wheels will be correct, and with `MQT_DEP=PIP` in `.travis.yml`, installation for test will be correct too.

When a constraint absolutely needs to be made on versions, it can be done in setup py ([as it is done for PyXB](https://github.com/OCA/l10n-italy/blob/7bc03a13543eb0e77bd38d7d0bd30882ba8dda8b/setup/l10n_it_fatturapa/setup.py#L8)). Also, please do avoid pinning dependencies when not strictly necessary (or pin a minimum version, or if nothing else works a maximum version). Strict pinning prevent your modules to work with the latest versions (including security updates) and makes it difficult to mix your modules in projects that may require more recent versions of the same dependencies.

- [x] 10.0: https://github.com/OCA/l10n-italy/pull/1702
- [ ] 11.0
 - [x] 12.0: https://github.com/OCA/l10n-italy/pull/1701

eLBati:
@sbidoul hi, fixing with https://github.com/OCA/l10n-italy/pull/1702 and https://github.com/OCA/l10n-italy/pull/1701
Thanks

About pyxb, we know the implications but the e-invoice modules  works with that version only.
We hope to get rid of pyxb in the future actually.

PR #1701, opened 2020-03-22
[12.0] IMP travis installation using dependencies from addon manifests

eLBati:
https://github.com/OCA/l10n-italy/issues/1700

--

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ addons:
 
 env:
   global:
-  - VERSION="12.0" TESTS="0" LINT_CHECK="0" MAKEPOT="0"
+  - VERSION="12.0" TESTS="0" LINT_CHECK="0" MAKEPOT="0" MQT_DEP=PIP
 
   matrix:
   - LINT_CHECK="1"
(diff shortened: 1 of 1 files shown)


eLBati:
Runbot seems to be red because of `The model intrastat.result.view has no _description` but that model has got `_description` actually (https://github.com/OCA/intrastat-extrastat/pull/69)

Then I noticed the pypi module is not up to date: https://pypi.org/project/odoo12-addon-intrastat-base/#history

Is runbot built with pypi packages for odoo modules?
@sbidoul do you want to check `odoo12-addon-intrastat-base`?

Thanks!

sbidoul:
Hi @eLBati,

So the issue is that the `intrastat_base` wheel has version 12.0.1.0.1 dating from May 2019 (without that `_description` field) and a fresher `12.0.1.0.0.dev...` with the fix but an older version number: You can see that in the [OCA wheelhouse](https://wheelhouse.odoo-community.org/oca-simple/odoo12-addon-intrastat-base/).

And travis (and thus runbot, via travis2docker) [correctly picks up the latest version](https://travis-ci.org/github/OCA/l10n-italy/jobs/665511421#L441-L442) (12.0.1.0.1).

Why it is so I don't know... the first hypothesis that comes to mind is that there was once that version 1.0.1 in the manifest and then some force push occured?

So I'll bum the version of `intrastat_base`.

sbidoul:
all green now

PR #1733, opened 2020-05-07
[12.0][FIX] to not use move name when it is equal to /

sergiocorato:
Vedi https://github.com/OCA/l10n-italy/issues/1734

--


primes2h:
@sergiocorato 
Vedi https://github.com/OCA/l10n-italy/pull/1736#issuecomment-625150931

PR #1910, opened 2020-10-08
[12.0] FIX fiscal_epos_print when setting debug info when order is not set

eLBati:
E.g. when getting error after sending commands like "aprire cassetto"

--


gigidn (approved):
it ok and safe to be merged

GSLabIt (approved):
lgtm

PR #1980, opened 2020-12-14
[12.0][IMP] l10n_it_account: Fix multi-locations installed

victoralmau:
Fix multi-locations installed

Locked by:

- [x] https://github.com/OCA/l10n-italy/pull/1981 (`l10n_it_fatturapa_pec`)

@Tecnativa TT26985

### DIFF
--- a/l10n_it_account/models/account_group.py
+++ b/l10n_it_account/models/account_group.py
@@ -53,11 +53,12 @@ class AccountGroup(models.Model):
 
         for progenitor in self.browse(tuple(set(progenitor_ids))):
             accounts = progenitor.get_group_accounts()
-            if not accounts.mapped('user_type_id').have_same_sign():
-                raise ValidationError(
-                    _("Incoherent balance signs for '{}' and its subgroups.")
-                    .format(progenitor.name_get()[0][-1])
-                )
+            if accounts and accounts[0].company_id.is_country_id_code_it():
+                if not accounts.mapped('user_type_id').have_same_sign():
+                    raise ValidationError(
+                        _("Incoherent balance signs for '{}' and its subgroups.")
+                        .format(progenitor.name_get()[0][-1])
+                    )
 
     @api.multi
     def _compute_account_balance_sign(self):
--- a/l10n_it_account/models/res_company.py
+++ b/l10n_it_account/models/res_company.py
@@ -1,0 +1,12 @@
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import models
+
+
+class ResCompany(models.Model):
+    _inherit = 'res.company'
+
+    def is_country_id_code_it(self):
+        self.ensure_one()
+        country_it = self.env.ref("base.it")
+        return self.partner_id.country_id == country_it
--- a/l10n_it_account/tests/test_l10n_it_account.py
+++ b/l10n_it_account/tests/test_l10n_it_account.py
@@ -13,18 +13,24 @@ class TestAccount(TransactionCase):
         self.group_1 = self.env['account.group'].create({
             'name': '1'
         })
+        self.company = self.env["res.company"].create({
+            "name": "Italy test company",
+            "country_id": self.env.ref("base.it").id,
+        })
 
     def test_group_constraint(self):
         self.env['account.account'].create({
             'name': 'it_account_1',
             'code': 'it_account_1',
             'group_id': self.group_1.id,
-            'user_type_id': self.data_account_type_current_assets.id
+            'user_type_id': self.data_account_type_current_assets.id,
+            'company_id': self.company.id
         })
         with self.assertRaises(ValidationError):
             self.env['account.account'].create({
                 'name': 'it_account_2',
                 'code': 'it_account_2',
                 'group_id': self.group_1.id,
-                'user_type_id': self.data_account_type_current_liabilities.id
+                'user_type_id': self.data_account_type_current_liabilities.id,
+                'company_id': self.company.id
             })
--- a/l10n_it_account/models/__init__.py
+++ b/l10n_it_account/models/__init__.py
@@ -4,3 +4,4 @@ from . import account_account
 from . import account_group
 from . import account_tax
 from . import account_type
+from . import res_company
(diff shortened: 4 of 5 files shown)


l10n_it_account/models/account_account.py
```diff
@@ -16,3 +16,10 @@ def check_balance_sign_coherence(self):
         # Avoid check upon empty recordset to make it faster
         if groups:
             groups.check_balance_sign_coherence()
+
+    def is_valid_l10n_it(self):
+        self.ensure_one()
+        it_ct = self.env.ref('l10n_it.l10n_it_chart_template_generic')
+        if not self.company_id or self.company_id.chart_template_id == it_ct:
+            return True
+        return False
```
pedrobaeza, line 25:
```suggestion
        return self.company_id or self.company_id.chart_template_id == it_ct
```

l10n_it_account/models/account_account.py
```diff
@@ -16,3 +16,10 @@ def check_balance_sign_coherence(self):
         # Avoid check upon empty recordset to make it faster
         if groups:
             groups.check_balance_sign_coherence()
+
+    def is_valid_l10n_it(self):
+        self.ensure_one()
```
pedrobaeza, line 21: Why not doing this at company level or account group level?

pedrobaeza:
@eLBati what do you think about this? We are implementing a subsidiary in Italy of an Spanish company and we have found this incompatibility.

FYI, we will migrate then corresponding modules to v13, as we want to migrate this customer to v13 and not wait to v14.

l10n_it_account/models/account_account.py
victoralmau: I have separated this logic in a function in case it needs to be used elsewhere and not duplicate code.
If you consider it unnecessary I will delete it.

pedrobaeza: I'm talking of adding this function at company level, the same as in l10n-spain

victoralmau: I don't find it, please copy link to example and will change it to `res.company`.

eLBati:
@pedrobaeza we have never put `l10n_it` as a dependency, the intent was to allow to load any COA module, even personalized.
I don't know if this is still a requirement, it is not in my customers' use cases.
@OCA/local-italy-maintainers what do you think?

If we wanted to keep that requirement, the `is_valid_l10n_it` check could be based on company's country...

sergiocorato:
@pedrobaeza @eLBati I don't install l10n_it on my clients either, because I use dedicated localizations.

eLBati:
I meant I always install `l10n_it`. Anyway what do you think about a check based on company's country?

victoralmau:
Travis CI error show error in test related to `l10n_it_fatturapa_in`
https://github.com/OCA/l10n-italy/pull/1983

sergiocorato:
> I meant I always install `l10n_it`. Anyway what do you think about a check based on company's country?

So whatever localization module is installed, seems better. And remove dependency too.

@victoralmau can you proceed about this?

victoralmau:
IMO `l10n_it_account` that extend account related to italian localization **is required to include `l10n_it in` dependens**. 
Now if everybody uset it `l10n_it` not exist any problem to add it, right?

A lot of addons depends to `l10n_it_account` and should implement with `TestL10nItAccountModBase` that include `l10n_it_chart_template_generic` (these required to add `l10n_it` dependency).

In the other hand and related to `is_valid_l10n_it`, now need l10n_it dependency but it's possible to check with `country_id.code` IT, what do you think about it? @pedrobaeza 

If keep on `l10n_it` dependency will be necessary to include in these PR all these changes: https://github.com/OCA/l10n-italy/pull/1983

Finally, all these change should to apply in v13, right?

sergiocorato:
@victoralmau As I don't always install `l10n_it` (as said before), many others can. Why put a not required dependency? I assure that an installation lives the same without :)

victoralmau:
> `l10n_it`

Ok, i will remove `l10n_it` dependency.

victoralmau:
Travis green now and changes applied. Ready to merge.

l10n_it_account/models/account_group.py
```diff
@@ -53,11 +53,12 @@ def check_balance_sign_coherence(self):
 
         for progenitor in self.browse(tuple(set(progenitor_ids))):
             accounts = progenitor.get_group_accounts()
-            if not accounts.mapped('user_type_id').have_same_sign():
-                raise ValidationError(
-                    _("Incoherent balance signs for '{}' and its subgroups.")
-                    .format(progenitor.name_get()[0][-1])
-                )
+            if accounts[0].company_id.is_country_id_code_it():
```
eLBati, line 56: Couldn't `accounts` be empty?

l10n_it_account/models/res_company.py
```diff
@@ -0,0 +1,11 @@
...
+
+from odoo import models
+
+
+class ResCompany(models.Model):
+    _inherit = 'res.company'
+
+    def is_country_id_code_it(self):
+        self.ensure_one()
+        return self.partner_id.country_id.code == 'IT'
```
pedrobaeza, line 11:
```suggestion
        country_it = self.env.ref("base.it")
        return self.partner_id.country_id == country_it
```

l10n_it_account/tests/test_l10n_it_account.py
```diff
@@ -4,6 +4,16 @@
...
 
+    def _company_create(self):
+        self.company = self.env['res.company'].create({
+            'name': 'Italy test company'
+        })
+        country_id_it = self.env['res.country'].search([
+            ('code', '=', 'IT')
+        ], limit=1)
+        self.company.partner_id.country_id = country_id_it.id
+        return True
```
pedrobaeza, line 15:
```suggestion
        return self.env["res.company"].create({
            "name": "Italy test company",
            "country_id": self.env.ref("base.it").id,
        })
```
and save the variable at the call line.

pedrobaeza (changes requested):

pedrobaeza (approved):

tafaRU (approved):

eLBati (approved):
Thanks

PR #2010, opened 2020-12-30
[12.0][IMP] l10n_it_fiscal_document_type: Hide fields to not italy company

victoralmau:

Locked by:
- [ ] `l10n_it_account` https://github.com/OCA/l10n-italy/pull/2011

@Tecnativa TT27569

### DIFF
--- a/l10n_it_fiscal_document_type/models/account_fiscal_position.py
+++ b/l10n_it_fiscal_document_type/models/account_fiscal_position.py
@@ -2,9 +2,11 @@ from odoo import models, fields
 
 
 class AccountFiscalPosition(models.Model):
-    _inherit = 'account.fiscal.position'
+    _name = 'account.fiscal.position'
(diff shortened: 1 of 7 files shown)


l10n_it_fiscal_document_type/models/account_fiscal_position.py
```diff
@@ -7,4 +7,22 @@ class AccountFiscalPosition(models.Model):
...
         'fiscal.document.type',
         string="Fiscal Document Type",
-        readonly=False)
+        readonly=False
+    )
+    is_company_it = fields.Boolean(
+        compute="_compute_is_company_it"
+    )
+
+    def _compute_is_company_it(self):
```
eLBati, line 16: As this check is always the same for the  3 models, could it be written only once?

eLBati: I noticed https://github.com/OCA/l10n-italy/pull/2014#issuecomment-752684726 later ;-)

PR #2011, opened 2020-12-30
[12.0][IMP] l10n_it_account: Hide fields to not italy company

victoralmau:

Locked by: 
- [ ] `l10n_multi_country` https://github.com/OCA/server-ux/pull/262

@Tecnativa TT27571

### DIFF
--- a/l10n_it_account/__manifest__.py
+++ b/l10n_it_account/__manifest__.py
@@ -18,6 +18,7 @@
         'account_fiscal_year',
         'account_tax_balance',
         'web',
+        'l10n_multi_country'
     ],
     "data": [
         'views/account_setting.xml',
--- a/l10n_it_account/models/account_group.py
+++ b/l10n_it_account/models/account_group.py
@@ -5,7 +5,8 @@ from odoo.exceptions import ValidationError
 
 
 class AccountGroup(models.Model):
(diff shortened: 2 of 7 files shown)


eLBati (approved):
LGTM

l10n_it_account/models/l10n_it_account_mixin.py
```diff
@@ -0,0 +1,19 @@
...
+    is_company_it = fields.Boolean(
+        compute="_compute_is_company_it"
+    )
+
+    def _compute_is_company_it(self):
+        for record in self:
+            if self.env.user.company_id.is_country_id_code_it():
+                record.is_company_it = True
+            else:
+                record.is_company_it = False
```
pedrobaeza, line 19: Simplify the expression, but also take into account the company of the object
```suggestion
            company = False
            if "company_id" in record._fields():
                company = record.company_id
            company = company or record.env.user.company_id
            record.is_company_it = company.is_country_id_code_it()
```

pedrobaeza (changes requested):
Can you develop a test for the mixin?

eLBati (changes requested):
A more generic solution is under consideration https://github.com/OCA/l10n-italy/pull/2017#pullrequestreview-561539307

PR #2012, opened 2020-12-30
[12.0][IMP] l10n_it_split_payment: Hide fields to not italy company

victoralmau:

Locked by:

@Tecnativa TT27568

### DIFF
--- a/l10n_it_split_payment/models/account.py
+++ b/l10n_it_split_payment/models/account.py
@@ -9,23 +9,27 @@ from odoo.exceptions import UserError
 
 
 class AccountFiscalPosition(models.Model):
-    _inherit = 'account.fiscal.position'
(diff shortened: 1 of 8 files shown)


l10n_it_split_payment/models/account.py
```diff
@@ -12,7 +12,24 @@ class AccountFiscalPosition(models.Model):
     _inherit = 'account.fiscal.position'
 
     split_payment = fields.Boolean('Split Payment')
+    is_company_it = fields.Boolean(
+        compute="_compute_is_company_it"
+    )
 
+    def _compute_is_company_it(self):
```
eLBati, line 19: https://github.com/OCA/l10n-italy/pull/2010#discussion_r550408833

PR #2014, opened 2020-12-30
[12.0][IMP] l10n_it_esigibilita_iva: Hide fields to not italy company

victoralmau:

Locked by:

@Tecnativa TT27566


pedrobaeza:
If there's a generic module that all depend on (I think it's l10n_it_account), you should create there a mixin and inherit from it on the rest of the models instead of having to add the same compute everywhere.

sergiocorato:
@victoralmau could you explain the functional reason for these changes?

pedrobaeza:

PR #2017, opened 2020-12-31
[12.0][IMP] l10n_it_ipa: Hide fields to not italy company

victoralmau:

Locked by:

@Tecnativa TT27566

### DIFF
--- a/l10n_it_ipa/model/l10n_it_ipa_mixin.py
+++ b/l10n_it_ipa/model/l10n_it_ipa_mixin.py
@@ -1,0 +1,27 @@
+# Copyright 2020 Tecnativa - Víctor Martínez
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+from odoo import models, fields
+
+
+class L10nItIpaMixin(models.AbstractModel):
+    _name = 'l10n_it_ipa.mixin'
+    _description = 'l10n_it_ipa Mixin'
+
+    is_company_it = fields.Boolean(
+        compute="_compute_is_company_it"
+    )
+
+    def _compute_is_company_it(self):
+        for record in self:
+            record.is_company_it = False
+            country_it = self.env.ref("base.it")
+            if (
+                (
+                    record.company_id and
+                    record.company_id.country_id == country_it
+                ) or (
+                    not record.company_id and
+                    self.env.user.company_id.country_id == country_it
+                )
+            ):
+                record.is_company_it = True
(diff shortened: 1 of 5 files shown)


l10n_it_ipa/model/l10n_it_ipa_mixin.py
```diff
@@ -0,0 +1,27 @@
+# Copyright 2020 Tecnativa - Víctor Martínez
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+from odoo import models, fields
+
+
+class L10nItIpaMixin(models.AbstractModel):
+    _name = 'l10n_it_ipa.mixin'
```
eLBati, line 7: Why another mixin in addition to `l10n_it_account.mixin`?

eLBati (changes requested):

l10n_it_ipa/model/l10n_it_ipa_mixin.py
victoralmau: Because this addon haven't `l10n_it_account` dependency.

eLBati: Right. Then maybe we should have a basic module (`l10n_it_multi_country`?) that introduces the mixin and make all these modules depend on it.

@pedrobaeza what do you think? Do you have anything similar in Spain?

@OCA/local-italy-maintainers cheddite?

eLBati: Probably this should designed at OCA level, not only for `l10n-italy`

pedrobaeza: Yeah, `l10n_it_base` containing the mixin is the best option, but not everyone will like to add on this version a new module dependency in the middle of the version (I would be one of them being on the opposite side).

pedrobaeza: Not bad idea, although in Spain we work at CoA level, not country code level. By the way, do you have a link to a complete Italian CoA that you use instead of the official one?

sergiocorato: I don't use this CoA for many customers, perhaps only some migrated, but for e.g.: https://github.com/sergiocorato/efatto/tree/12.0/l10n_it_generic
I agree with a module `l10n_multi_country` at OCA level, to add as dependency in modules of whatever localization, or to specific newly created modules inheriting the standard modules.

PR #2019, opened 2020-12-31
[12.0][IMP] l10n_it_rea: Hide fields to not italy company

victoralmau:

Locked by:

@Tecnativa TT27566

### DIFF
--- a/l10n_it_rea/models/l10n_it_rea_mixin.py
+++ b/l10n_it_rea/models/l10n_it_rea_mixin.py
@@ -1,0 +1,27 @@
+# Copyright 2020 Tecnativa - Víctor Martínez
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+from odoo import models, fields
+
+
+class L10nItReaMixin(models.AbstractModel):
+    _name = 'l10n_it_rea.mixin'
+    _description = 'l10n_it_rea Mixin'
+
+    is_company_it = fields.Boolean(
+        compute="_compute_is_company_it"
+    )
+
+    def _compute_is_company_it(self):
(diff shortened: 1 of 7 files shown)


l10n_it_rea/models/l10n_it_rea_mixin.py
```diff
@@ -0,0 +1,27 @@
+# Copyright 2020 Tecnativa - Víctor Martínez
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+from odoo import models, fields
+
+
+class L10nItReaMixin(models.AbstractModel):
+    _name = 'l10n_it_rea.mixin'
```
eLBati, line 7: https://github.com/OCA/l10n-italy/pull/2017#pullrequestreview-561539307

victoralmau: Because this addon haven't `l10n_it_account` dependency.

pedrobaeza: @eLBati is it OK to add such dependency on the modules instead?

eLBati: https://github.com/OCA/l10n-italy/pull/2017#discussion_r551947927

I think it's better  not to add `l10n_it_account` as a dependency of modules like  `l10n_it_rea`, because some database could use these basic modules without using accounting.

pedrobaeza: OK, I see, then maybe a `l10n_it_base` containing this mixin can be an option, or to duplicate code.

PR #2051, opened 2021-01-18
[12.0][FIX] l10n_it_fatturapa_out: Fix error tests to copy files

victoralmau:
Fix error tests to copy files

Please @pedrobaeza review it

@Tecnativa TT26985

### DIFF
--- a/l10n_it_fatturapa_out/readme/CONTRIBUTORS.rst
+++ b/l10n_it_fatturapa_out/readme/CONTRIBUTORS.rst
@@ -4,3 +4,7 @@
 * Alessio Gerace
 * Alex Comba
 * Sergio Zanchetta <https://github.com/primes2h>
+
+* `Tecnativa <https://www.tecnativa.com>`_:
+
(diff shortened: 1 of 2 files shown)


eLBati:
When does the error happen?

pedrobaeza:
And why a copy of the file is needed?

victoralmau:
Errors happen to run tests.
I don't understand code (reason to copy file and why only error happen only in some case) but it's my fast fix it.

pedrobaeza:
@eLBati can you clarify what is the goal of such copy to see if this can be improved in any way?

eLBati:
:thinking: I found the commit which introduced it ec69c24a0054b9a1047ed58aae04dbbf34f8e5fe but I don't remember the reason.
Maybe @archetipo ?

Anyway, if tests run correctly without the copy, I guess we can directly remove it

eLBati (approved):
Thanks

pedrobaeza (approved):

PR #2052, opened 2021-01-18
[12.0][FIX] l10n_it_fatturapa_in: Fix error tests in previous chart template load (multi-location)

victoralmau:
Tests about it work fine but if is installed with other addons (`l10n_es_vat_book` for example) appear errors in test because other addon load account chart template and not exist accounts related to some types.
Although isn't a very good solution it's a fast fix.

IMO a good solution it's very difficult because will need to use a base test that auto-install itally account chart template and prevent errors in tests BUT affect to almost all addons to `l10n-italy`.

@Tecnativa TT26985

### DIFF
--- a/l10n_it_fatturapa_in/readme/CONTRIBUTORS.rst
+++ b/l10n_it_fatturapa_in/readme/CONTRIBUTORS.rst
@@ -4,3 +4,7 @@
 * Sergio Zanchetta <https://github.com/primes2h>
 * Giovanni Serra <giovanni@gslab.it>
 * Gianmarco Conte <gconte@dinamicheaziendali.it>
+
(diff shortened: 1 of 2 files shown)


pedrobaeza (commented):
The main problem here is that this is using current company and chart of accounts (CoA) without taking into account which is the CoA and for which country. This should be used a similar technique like `l10n_es_aeat` that creates a company, and load Spanish CoA for using it.

victoralmau:
Speak about it in othe PR https://github.com/OCA/l10n-italy/pull/2017#pullrequestreview-561539307 related to https://github.com/OCA/l10n-italy/issues/2053

PR #2245, opened 2021-04-27
[8.0][FIX] format works in ASCII

### LINKED ISSUE #2244: [8.0] Generazione Comunicazione dati fattura non creata per errore
La funzione `format()` di python 2.7 ha un baco vedi [1] che genera questo [2] errore qui [3] in caso di stringa con caratteri particolari come [4]. Non genera problemi ad esempio con caratteri non latini.
La funzione `%` invece non genera questo problema.

[1] https://stackoverflow.com/questions/54364991/ascii-decode-error-even-though-everything-being-unicode-python-2-7
[2] `UnicodeEncodeError: 'ascii' codec can't encode character u'\xe9' in position 7: ordinal not in range(128)`
[3] https://github.com/OCA/l10n-italy/blob/493acc67c5a93d4359a5d6078d714bca1a2918f8/l10n_it_invoices_data_communication/models/communication.py#L290
[4] `98, Allée des Champs Elysées `

sergiocorato:
fix #2244

### DIFF
--- a/l10n_it_invoices_data_communication/models/communication.py
+++ b/l10n_it_invoices_data_communication/models/communication.py
@@ -287,9 +287,9 @@ class ComunicazioneDatiIva(models.Model):
             partner.commercial_partner_id.fiscalcode or ''
         vals['cedente_Denominazione'] = encode_for_export(partner.name, 80)
         # Sede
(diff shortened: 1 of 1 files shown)


sergiocorato:
Close for inactivity

PR #2294, opened 2021-05-21
[10.0][FIX] l10n_it_intrastat_statement: la rettifica può essere legata a un periodo di tipo diverso

SimoRubi:
https://github.com/OCA/l10n-italy/issues/2293 per `10.0`

--

### DIFF
--- a/l10n_it_intrastat_statement/tests/test_intrastat_statement.py
+++ b/l10n_it_intrastat_statement/tests/test_intrastat_statement.py
@@ -163,6 +163,30 @@ class TestIntrastatStatement (AccountingTestCase):
         # Last line is section line, for quarter report it should be 64 chars
         self.assertEqual(len(file_content.splitlines()[-1]), 64)
 
+    def test_statement_sale_quarter_adjustment_period(self):
+        """Check that sale adjustments can be in a different period type"""
+        invoice = self._get_intrastat_computed_invoice()
+        month = fields.Date.from_string(invoice.date_invoice).month
+        quarter = 1 + (month - 1) // 3
+        statement = self.statement_model.create({
+            'period_number': quarter,
+            'period_type': 'T',
+        })
+        statement.compute_statement()
+
+        adjustment_vat_code = 'adjust_vat'
+        statement.sale_section2_ids = [(0, 0, {
+            'partner_id': invoice.partner_id.id,
+            'vat_code': adjustment_vat_code,
+            'sign_variation': '+',
+            'year_id': fields.Date.from_string(invoice.date_invoice).year - 1,
+            'month': 1,
+        })]
+        file_content = statement \
+            .with_context(sale=True) \
+            .generate_file_export()
+        self.assertIn(adjustment_vat_code, file_content)
+
     def test_statement_purchase(self):
         bill = self._get_intrastat_computed_bill()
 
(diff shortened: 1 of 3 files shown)


l10n_it_intrastat_statement/tests/test_intrastat_statement.py
```diff
@@ -163,6 +163,30 @@ def test_statement_sale_quarter(self):
         # Last line is section line, for quarter report it should be 64 chars
         self.assertEqual(len(file_content.splitlines()[-1]), 64)
 
+    def test_statement_sale_quarter_adjustment_period(self):
+        """Check that sale adjustments can be in a different period type"""
+        invoice = self._get_intrastat_computed_invoice()
+        month = fields.Date.from_string(invoice.date_invoice).month
+        quarter = 1 + (month - 1) // 3
```
sergiocorato, line 170: Operator `//` in python2 exists?

l10n_it_intrastat_statement/tests/test_intrastat_statement.py
```diff
@@ -163,6 +163,30 @@ def test_statement_sale_quarter(self):
...
+        })
+        statement.compute_statement()
+
+        adjustment_vat_code = 'adjust_vat'
+        statement.sale_section2_ids = [(0, 0, {
+            'partner_id': invoice.partner_id.id,
+            'vat_code': adjustment_vat_code,
+            'sign_variation': '+',
+            'year_id': fields.Date.from_string(invoice.date_invoice).year,
+            'month': 1,
```
sergiocorato, line 183: `month` perhaps?

l10n_it_intrastat_statement/tests/test_intrastat_statement.py
OpenCode: Yes: https://docs.python.org/2.7/reference/lexical_analysis.html#operators

l10n_it_intrastat_statement/tests/test_intrastat_statement.py
SimoRubi: Here I am testing that the adjustment can be in a different period type than the `statement` so it doesn't have to be related to the `statement`'s month.
`month` is only used to compute the `quarter` of the `statement`.

l10n_it_intrastat_statement/tests/test_intrastat_statement.py
sergiocorato: Thanks

l10n_it_intrastat_statement/tests/test_intrastat_statement.py
sergiocorato: In this way (putting always 1 as `month`), in the first three month of the year month would be in the same quarter of statement, is it wanted?

SimoRubi: The code now allows the period stated in the adjustment to be independent from the period of the `statement` and this is what this test case is checking.
Anyway, I understand that this wouldn't be a realistic use-case when they are overlapping, so I changed the adjustment's year; can you check?

BTW, thanks for the review :)

sergiocorato: Ok (I don't use this version anyway)

sergiocorato (approved):

SimoRubi:
> What a great day to merge this nice PR. Let's do it!
> Prepared branch [10.0-ocabot-merge-pr-2294-by-sergiocorato-bump-patch](https://github.com/OCA/l10n-italy/commits/10.0-ocabot-merge-pr-2294-by-sergiocorato-bump-patch), awaiting test results.

Credo si sia un po' piantato, @sergiocorato posso fare qualcosa?

sergiocorato:
vero, vediamo se serve a qualcosa:
/ocabot merge patch

PR #2392, opened 2021-08-13
[14.0][FIX] l10n_it_fatturapa_out: (multicompany) choose the right company

TheMule71:
Fix for: #2391 

We need to use company_id from the invoice.moves and not from the wizard environment. In 14.0 there can be multiple companies selected at the same time.

--


francescapenso:
testata e funzionante

PR #2436, opened 2021-09-24
[FIX] l10n_it_fiscal_document_type: license must be AGPL since it depends on AGPL modules

dreispt:
image


simahawk (commented):
@alessandrocamilli @eLBati can you check if it is ok for you and if this affects other modules?

eLBati:
@simahawk it should be OK. Ref https://github.com/OCA/l10n-italy/pull/2360#issuecomment-885468210

eLBati (approved):
@dreispt thanks!

scigghia (approved):

PR #2624, opened 2022-01-31
[REF] l10n_it_withholding_tax: Bump version

SimoRubi:
Currently in OCA wheelhouse modules for `l10n_it_withholding_tax` there is this situation:
image
(from https://wheelhouse.odoo-community.org/oca-simple/odoo14-addon-l10n-it-withholding-tax/)

The problem is that the most advanced version is [14.0.1.1.2.dev1](https://wheelhouse.odoo-community.org/oca-simple/odoo14-addon-l10n-it-withholding-tax/odoo14_addon_l10n_it_withholding_tax-14.0.1.1.2.dev1-py3-none-any.whl) but it is not the newest: it has been generated in November and does not include the latest merged changes for the module (https://github.com/OCA/l10n-italy/pull/2617).
These changes are instead contained in [14.0.1.1.2](https://wheelhouse.odoo-community.org/oca-simple/odoo14-addon-l10n-it-withholding-tax/odoo14_addon_l10n_it_withholding_tax-14.0.1.1.2-py3-none-any.whl).

As a solution, in this PR I am simply bumping the module version, but before marking as "Ready for merge" it might be worth to investigate the issue further, @sbidoul could you please have a look?

### DIFF
--- a/l10n_it_withholding_tax/__manifest__.py
+++ b/l10n_it_withholding_tax/__manifest__.py
@@ -5,7 +5,7 @@
 
 {
     "name": "ITA - Ritenute d'acconto",
-    "version": "14.0.1.1.2",
+    "version": "14.0.1.1.3",
     "category": "Account",
     "author": "Openforce, Odoo Italia Network, " "Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/l10n-italy",


sbidoul:
@SimoRubi 

```
>>> from packaging.version import Version
>>> Version("14.0.1.1.2.dev1") < Version("14.0.1.1.2")
True
>>> 
```

So 14.0.1.1.2 is the latest one, that sounds correct to me.

SimoRubi:
> @SimoRubi
> 
> ```
> >>> from packaging.version import Version
> >>> Version("14.0.1.1.2.dev1") < Version("14.0.1.1.2")
> True
> >>> 
> ```
> 
> So 14.0.1.1.2 is the latest one, that sounds correct to me.

I always assumed `1.0.dev1` was `1.0` + some development, instead looking in https://www.python.org/dev/peps/pep-0440 it actually means something like **One development toward `1.0`** so it precedes `1.0`.
The ordering in https://wheelhouse.odoo-community.org/oca-simple/odoo14-addon-l10n-it-withholding-tax/ is also misleading and now I assume it is simply alphabetical.

Thanks for the clarification and sorry for the silly question :)

sbidoul:
Yes the ordering on wheelhouse.acsone.eu is alphabetical and misleading.
On PyPI it is clearer: https://pypi.org/project/odoo14-addon-l10n-it-withholding-tax/#history

PR #2632, opened 2022-02-04
[14.0][FIX] l10n_it_fatturapa_out_rc: skip preventive checks when there are no invoices with rc

tafaRU:
Fix https://github.com/OCA/l10n-italy/issues/2631


TheMule71 (approved):
LGTM

PR #2753, opened 2022-04-08
[FIX] l10n_it_reverse_charge: disable one test (for now)

TheMule71:
https://github.com/OCA/OCB/commit/674eaa3bc10210d34a02145529516da1fa1c1b66
apparently broke odoo.addons.account.tests.common.AccountTestInvoicingCommon
method init_invoice() when called with two opposed values as amounts
(throws odoo.exceptions.UserError: Cannot create unbalanced journal
entry)

One test (test_intra_EU_zero_total) disabled. If/when init_invoice() is
fixed, it will trigger a failure, or the test will be rewritten to adapt
to the new behaviour, whichever comes first.

### DIFF
--- a/l10n_it_fatturapa_out/i18n/am.po
+++ b/l10n_it_fatturapa_out/i18n/am.po
@@ -302,6 +302,11 @@ msgid "Followers (Partners)"
 msgstr ""
 
 #. module: l10n_it_fatturapa_out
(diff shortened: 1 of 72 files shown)


eLBati (approved):

eLBati:
PS: aggiungendo un link alla issue su odoo

SirTakobi:
Issue di riferimento: https://github.com/odoo/odoo/issues/88858

PR #2763, opened 2022-04-15
[14.0] fix test tag post-install

TheMule71:
After 
https://github.com/odoo/odoo/commit/81aac30dd2278e43a43cf8fd6cef31e1e8c60f3f
tests that depend on having a coa installed must be tagged 'post-install'

(from #2753)
https://github.com/OCA/OCB/commit/674eaa3bc10210d34a02145529516da1fa1c1b66
breaks init_invoice()


TheMule71:
opps, correggo

PR #2768, opened 2022-04-22
[IMP] l10n_it_fatturapa_out: Fixed PyXB version

SirTakobi:
Fix error:
ERROR: Double requirement given: PyXB==1.2.6 (from -r test-requirements.txt (line 17)) (already in pyxb (from -r test-requirements.txt (line 16)), name='PyXB')
during setuptools-odoo external dependencies retrieval

### DIFF
--- a/setup/l10n_it_fatturapa_out/setup.py
+++ b/setup/l10n_it_fatturapa_out/setup.py
@@ -5,6 +5,7 @@ setuptools.setup(
     odoo_addon={
         'external_dependencies_override': {
             'python': {
+                'pyxb': 'PyXB==1.2.6',
                 'unidecode': 'Unidecode<1.3.0',
             },
         },


sbidoul:
You need to pin pyxb in `setup/l10n_it_fatturapa_out/setup.py` like it is done for l10n_it_fatturapa.
Otherwise the old pip version used for python 2 can't resolve by itself.

SirTakobi:
> You need to pin pyxb in `setup/l10n_it_fatturapa_out/setup.py` like it is done for l10n_it_fatturapa. 

Thanks, I hadn't noticed that there was the same requirement in **l10n_it_fatturapa_out** too.
I did as you suggested, and it is working! :tada:

> Otherwise the old pip version used for python 2 can't resolve by itself.

Maybe a requirements file like
```
PyXB
PyXB==1.2.6
```
should simply install version `1.2.6`, but this is `pip`'s behaviour and I can actually see why they prefer to raise an error in this situation.

sbidoul:
> should simply install version 1.2.6, but this is pip's behaviour and I can actually see why they prefer to raise an error in this situation.

That is what recent pip version do indeed, but the python 2 version of pip is frozen forever.

SirTakobi:
@OCA/local-italy-maintainers che ne dite si può mergiare o devo modificare qualcosa? Senza questa modifica il runboat di `10.0` è :x:

sergiocorato (approved):

matteoopenf (approved):

PR #2777, opened 2022-05-02
[12.0] FIX l10n_it_fatturapa_in, Setting Italian timezone by default, otherwise e_invoice_received_date could be set the day before of the actual date

eLBati:
https://github.com/OCA/l10n-italy/issues/2778

Example:
receiving date = 2022-04-01 00:16
saved in DB, UTC format, as 2022-03-31 22:16:03
converted to date = 2022-03-31

In this case, the invoice would be registered in the wrong period (month)


sergiocorato (approved):
LGTM

mrcast (approved):
Functional test ok

tafaRU (approved):

eLBati:
@tafaRU fatto

PR #2800, opened 2022-05-15
Update post-migration.py

OmniaGit:
in case you do not have any data in the line_ids the script fail
with this correction the problem is solved


tafaRU:
Chiudo per inattività in favore di https://github.com/OCA/l10n-italy/pull/2826.

PR #2925, opened 2022-08-31
[12.0] FIX l10n_it_sdi_channel - ValueError: Wrong value for ir.attachment.type: 'out_invoice'

eLBati:
When "Validate, export and send to SdI"


TheMule71 (approved):
LGTM

PR #2969, opened 2022-10-04
[12.0][FIX] - l10n_it_sdi_channel : user error for unset ES channel in a company record

BojanOD:
- Informing the user what actually went wrong.
- Instead of showing the message _ValueError: Expected singleton: sdi.channel()_


eLBati:
Hi @BojanOD could you describe the steps to reproduce the problem on a new DB?
Thanks

PR #3023, opened 2022-11-10
[14.0][IMP]l10n_it_abicab: Remove unnecessary method "onchange_bank_id"

TonyMasciI:
Removed method because there is a "related" attribute in fields bank_abi & bank_cab which makes the method unnecessary


Borruso (approved):
LGTM

eLBati (approved):
Grazie

PR #3059, opened 2022-11-22
[14.0] [FIX] l10n_it_declaration_of_intent: avoid duplicating messages in in…

primes2h:
…voices' notes when an invoice is set back to draft and then validated again

Risolve https://github.com/OCA/l10n-italy/issues/2328 per la versione 14.0


marcelofrare (approved):
runboat is ok

PR #3076, opened 2022-11-29
[12.0] FIX fiscal_epos_print: always set beginFiscalReceipt, after REFUND message, in case of refunds

eLBati:


eLBati:
https://discord.com/channels/753902328494424064/776479965411541004/1048254976041963561

PR #3089, opened 2022-12-16
[12.0][FIX]l10n_it_sdi_channel: restore python 3.5 compatibility

aleuffre:
The trailing comma in the arguments of methods such as this one

```python
def _prepare_attachment_in_values(
       self,
       file_name,
       file_content,
       **default_values,
   ):
```

breaks compatibility with python 3.5. Support for this syntax [was added in Python 3.6](https://bugs.python.org/issue9232). It's sufficient to remove the trailing comma after `default_values` in the 3 methods affected.

### DIFF
--- a/l10n_it_sdi_channel/models/sdi.py
+++ b/l10n_it_sdi_channel/models/sdi.py
@@ -63,7 +63,7 @@ class SdiChannel(models.Model):
         self,
         file_name,
         file_content,
-        **default_values,
+        **default_values
     ):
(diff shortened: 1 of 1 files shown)


SirTakobi:
Thanks, the CI should have caught that, I have proposed the Python version change in https://github.com/OCA/oca-ci/pull/39, let's see how it goes

sergiocorato (approved):

PR #3092, opened 2022-12-19
[10.0][FIX] - 10n_it_reverse_charge: wouldn't let you set rc invoice to draft

michelerusti:
because context parameter was not set.

Corregge #3093  per 10.0


stevech091 (approved):
Functional test OK

PR #3281, opened 2023-04-19
[FIX] assets_management: correct permission management

PicchiSeba:
The previous version allowed `base.group_user` to access the assets with full read/write permissions from their url even if they were hidden from them in the menus.
With this PR we aim to correctly manage user permissions, already implemented but never used.

Also, child menu items were being used with redundant group permission already specified by the parent.


aleuffre (approved):

aleuffre:
Avrei proposto anche due livelli ma per ora partiamo così, meglio della situazione precedente.

sergiocorato (approved):
LGTM

francesco-ooops:
@PicchiSeba puoi far ricostruire il runboat?

PR #3513, opened 2023-08-03
[IMP] l10n_it_fatturapa_in: Update logic to import e-invoice xml

patrickt-oforce:
Update logic to import e-invoice xml, porting logic from PR #3100 and adapting it to v. 14.0


matteoopenf:
Ciao @monen17  questa e' un porting della tua pr a 14, per caso hai qualche suggerimento?
Grazie Mille

PR #3558, opened 2023-09-04
[16.0][IMP] l10n_it_delivery_note: remove misleading price_unit sum total

aleuffre:
Removing the price_unit sum from the list of products in the delivery note since the sum of price_unit is not particularly meaningful and can be misleading (it doesn't take quantity into account)

Cf: #3557


sergiocorato (approved):

andreampiovesana (approved):
LGTM

FrancescoBallerini:
Chiedo scusa avevo referenziato questa in un commit invece di 3588

PR #3611, opened 2023-09-27
[IMP] l10n_it_delivery_note: add hook to prepare vals needed to create DN

tafaRU:


tafaRU:
@Borruso potresti fare review?

PR #3627, opened 2023-10-03
[IMP][l10n_it_asset_history_import] add supplier in import issue http…

matteoopenf:
…s://github.com/OCA/l10n-italy/issues/3625

### DIFF
--- a/l10n_it_asset_history_import/wizards/asset_history_import.py
+++ b/l10n_it_asset_history_import/wizards/asset_history_import.py
@@ -461,7 +461,7 @@ class AssetHistoryImport(models.TransientModel):
             partner = partner.search([("vat", "ilike", supplier_vat)])[:1]
             if not partner:
                 raise ValidationError(
-                    _("Could not find Supplier with VAT {}".format(supplier_vat))
+                    _("Could not find Supplier with VAT %s", supplier_vat)
                 )
 
         supplier_doc_ref_col = get_header_by_model_and_field(
--- a/l10n_it_asset_history_import/wizards/asset_history_import.py
+++ b/l10n_it_asset_history_import/wizards/asset_history_import.py
@@ -251,6 +251,26 @@ HEADERS = (
         tmpl_default=0,
         type="float",
     ),
+    Header(
+        name=_("Supplier VAT"),
+        col=22,
+        model="res.partner",
+        field="vat",
+        apply_on="asset.asset",
+        required=False,
+        tmpl_default="",
+        type="str",
+    ),
+    Header(
+        name=_("Supplier Doc Ref"),
+        col=23,
+        model="asset.asset",
+        field="supplier_ref",
+        apply_on="asset.asset",
+        required=False,
+        tmpl_default="",
+        type="str",
+    ),
 )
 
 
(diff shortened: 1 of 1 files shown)


l10n_it_asset_history_import/wizards/asset_history_import.py
```diff
@@ -434,15 +454,29 @@ def convert_to_asset_vals(self, asset_dict, workbook, sheet):
                     "Is it in its ISO 4217 format?"
                 )
             )
+        supplier_vat_col = get_header_by_model_and_field("res.partner", "vat").col
+        supplier_vat = asset_dict[supplier_vat_col]
+        partner = self.env["res.partner"]
+        if supplier_vat:
+            partner = partner.search([("vat", "ilike", supplier_vat)])[:1]
```
SilvioC2C, line 461: Might be better for performances to use `limit=1` as `search` parameter instead of `[:1]`.
Else, you can use `fields.first(partner.search([("vat", "ilike", supplier_vat)]))` which will also guarantee to maintain the recordset order.

l10n_it_asset_history_import/wizards/asset_history_import.py
```diff
@@ -434,15 +454,29 @@ def convert_to_asset_vals(self, asset_dict, workbook, sheet):
...
+            if not partner:
+                raise ValidationError(
+                    _("Could not find Supplier with VAT {}".format(supplier_vat))
```
SilvioC2C, line 464: Since Odoo 14.0, the `_()` allows usage of placeholders: https://github.com/odoo/odoo/commit/3c3b88f6768e31417263b3828a874ca204e29514

It's good practice to use it when possible:
```suggestion
                    _("Could not find Supplier with VAT %s", supplier_vat)
```

SilvioC2C (approved):
LGTM overall, some non-blocking comments

l10n_it_asset_history_import/wizards/asset_history_import.py
GSLabIt, line 461:
```suggestion
            partner = partner.search([("vat", "ilike", supplier_vat)], limit=1)
```

l10n_it_asset_history_import/wizards/asset_history_import.py
```diff
@@ -434,15 +454,29 @@ def convert_to_asset_vals(self, asset_dict, workbook, sheet):
...
+        supplier_doc_ref_col = get_header_by_model_and_field(
+            "asset.asset", "supplier_ref"
+        ).col
+        supplier_doc_ref = asset_dict[supplier_doc_ref_col]
         vals.update(
             {
                 "category_id": categ.id,
                 "company_id": self.company_id.id,
                 "currency_id": curr.id,
+                "supplier_id": partner and partner.id or False,
```
SilvioC2C, line 476: Don't need to check if the partner has been found:
1. since we defined `partner = self.env["res.partner"]`, then `partner.id` will be `False` anyway if no partner was found
2. also, raising the `ValidationError` if no partner is found, makes sure that we reach this part of the code only if `partner` is correctly valued
```suggestion
                "supplier_id": partner.id,
```

francesco-ooops:
@matteoopenf riuscite a completarla?

matteoopenf:
> @matteoopenf riuscite a completarla?

subito no

PR #3662, opened 2023-10-13
[16] change of mainteiner name on manifest

MarcoCalcagni:


TheMule71:
vd https://github.com/OCA/l10n-italy/pull/3663

PR #3684, opened 2023-10-27
[14.0] [FIX] github pre-commit action: pin python version to 3.11 in repository

renda-dev:
As previously stated in https://github.com/OCA/oca-addons-repo-template/pull/214, I'm bringing this change to this repo in version 14.0

CC: @OCA/local-italy-developers, @pedrobaeza

PR #3685, opened 2023-10-27
[14.0] dotfiles update needs manual intervention

etobella:
Dear maintainer,

After updating the dotfiles, `pre-commit run -a`
fails in a manner that cannot be resolved automatically.

Can you please have a look, fix and merge?

Thanks,


aleuffre:
Starting to work on this

matteoopenf:
but the pre-commit not works

aleuffre:
> but the pre-commit not works

precisely, I'm working to fix the pre-commit issues

aleuffre:
But pre-commit is working, it's just that new checks were added. It's the same deal in every repo.

If you check the pre-commit logs, it's only failing because the modules are not up to standard, and often use deprecated syntax or dangerous functions such as `position="replace"`

This PR has been generated by the mass update of all repos from https://github.com/OCA/oca-addons-repo-template/ . #3684 just manually applied one of the changes as a stopgap, but this contains that change and more. This one will have to be merged eventually, in order to stay up to date with the rest of OCA.

PR #3686, opened 2023-10-27
[16.0] dotfiles update needs manual intervention

etobella:
Dear maintainer,

After updating the dotfiles, `pre-commit run -a`
fails in a manner that cannot be resolved automatically.

Can you please have a look, fix and merge?

Thanks,


SirAionTech:
I think we can close this one: we can continue in https://github.com/OCA/l10n-italy/pull/3767 where I have updated the copier template to the latest version.

PR #3690, opened 2023-10-31
14.0 - l10n_it_central_journal - Keep last start row when new daterange start row value is zero

patrickt-oforce:


matteoopenf (approved):
LGTM

l10n_it_central_journal/wizard/print_giornale.py
```diff
@@ -80,7 +80,7 @@ def on_change_daterange_id(self):
             self.date_move_line_to = date_end
             if self.daterange_id.progressive_line_number != 0:
                 self.start_row = self.daterange_id.progressive_line_number + 1
-            else:
+            if self.daterange_id.progressive_line_number == 0 and not self.start_row:
```
sergiocorato, line 83:
```suggestion
            elif not self.start_row:
```

PR #3734, opened 2023-11-23
[14.0] [FIX] l10n_it_fatturapa_in_rc: Fix roundings in error check

patrickt-oforce:
When import / autocomplete the invoice with RC lines, the rounding of amount total has 3 decimal position, and using `float_round` method of Odoo round total amount to next decimal value; using integers allow to obtain correct value for rounding


francesco-ooops:
@patrickt-oforce hai idea del perchè i test falliscono?

PR #3746, opened 2023-11-30
[14.0]l10n_it_riba_sale_commission: Partial commission set on Riba partial payments

odooNextev:


l10n_it_riba_sale_commission/models/account_move.py
```diff
@@ -22,11 +22,98 @@ def _skip_future_payments(self, date_payment_to):
...
+                        default=0,
                     )
                 )
                 for line in self.invoice_id.line_ids.filtered(lambda r: r.date_maturity)
             ]
             if dates:
                 return date_payment_to < max(dates)
         return super()._skip_future_payments(date_payment_to)
+
+    def _partial_commissions(self, date_payment_to):
```
renda-dev, line 35: Instead of rewriting the entire method, wouldn't be possible to inherit `_get_reconciled_invoices_partials` and remove the lines that you are excluding here?

odooNextev: Hi @renda-dev, Thanks for your review. I inherited the method _get_reconciled_invoices_partials and excluded partial payments directly there.

l10n_it_riba_sale_commission/models/account_move.py
```diff
@@ -10,6 +10,27 @@ class AccountMove(models.Model):
...
+                [("move_line_id", "=", partial.debit_move_id.id)]
+            )
+            if riba_mv_line:
+                riba_type = riba_mv_line.riba_line_id.type
+                if (
+                    riba_mv_line.riba_line_id.due_date
+                    + timedelta(days=riba_mv_line.riba_line_id.config_id.safety_days)
+                    > date.today()
+                ) and riba_type == "sbf":
+                    to_remove.append((partial, amount, counterpart_line))
```
renda-dev, line 31: Looking at this in the future could possibly be pretty confusing, could you please add a comment inside this method, wherever do you prefer, explaining why you are ignoring this lines? Thanks!

renda-dev (changes requested):
Could you add some tests to make sure that only the right lines are ignored?

And please, put `[FIX]` in the MR title as requested in the [contributing docs](https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#commit-message)

renda-dev (approved):
Code review, LGTM

francesco-ooops (approved):
Functional ok!

PR #3748, opened 2023-11-30
[14.0][FIX] l10n_it_withholding_tax add a field with amount net pay signed to show in tree view

sergiocorato:


francesco-ooops:
@sergiocorato ci siamo capiti :)

PR #3753, opened 2023-12-05
[14.0][IMP] add ability to issue positive with negative riba checking total is always positive

sergiocorato:


francesco-ooops:
@sergiocorato issue?

PR #3770, opened 2023-12-07
[14.0][FIX] l10n_it_delivery_note_inter_company: multiple pickings on DN error

renda-dev:
Fix di un errore che sarebbe comparso quando si tentava di aggiungere un picking a un DN già creato.


l10n_it_delivery_note_inter_company/models/stock_delivery_note.py
```diff
@@ -27,13 +28,13 @@ def update_detail_lines(self):
             [("delivery_note_id", "!=", False)]
         )
```
aleuffre, line 29: **issue(non-blocking)**: This is potentially A LOT of pickings that you're loading in memory.

I marked it as non-blocking just because it's not technically part of thsis PR, but since the module is still new, I think it's a good time to fix it.

Also, a few lines below, you then only want to operate on pickings that have the `intercompany_picking_id` field set, so you could already add that condition here. Ideally, though, I think you'd want to do a `search` in the `for` loop instead of loading all possible pickings here.

Or just use `note.picking_ids`.

aleuffre (changes requested):
partial code review, a few structural concerns

l10n_it_delivery_note_inter_company/models/stock_delivery_note.py
renda-dev: Thanks for the beautiful suggestion, I'm still not totally sure why in the past I wrote that superficial `filtered`, but using `note.picking_ids` totally works out.

aleuffre (approved):
Code review, LGTM

francesco-ooops (approved):
Funzionale ok!

francesco-ooops:
@tafaRU 🙏?

francesco-ooops:
Issue non necessaria finchè il modulo non viene portato a v16

@MarcoCalcagni se riesci, ci sarebbe molto utile mergiare questo fix :)

renda-dev:
df037bcb9db08521d1d3a74a0e241deafea5e9a8 moved to https://github.com/OCA/l10n-italy/pull/3789

PR #3965, opened 2024-02-15
[16.0] FIX l10n_it_intrastat: multi company for account.invoice.intrastat

eLBati:
Otherwise, going to Contabilità > Righe intrastat, user would see data of every company

PR #3997, opened 2024-02-27
FIX l10n_it_intrastat: multi company for account.invoice.intrastat

sergiocorato:
Otherwise, going to Contabilità > Righe intrastat, user would see data of every company


sergiocorato:
Backport di #3965

robyf70 (commented):
Potresti aggiungere il **company_id** anche nelle varie view inclusa la search?

PR #4019, opened 2024-03-06
[FIX][l10n_it_reverse_charge] fix error when supplier invoice is anot…

matteoopenf:
Risolve https://github.com/OCA/l10n-italy/issues/4020


amosdisabatino (approved):

sergiocorato (approved):
Grazie

PR #4057, opened 2024-03-15
[14.0] [IMP] l10n_it_delivery_note: prevent invoice creation

renda-dev:
In some cases, the possibility to create an invoice for a single DN could be misleading.
- Added "Do not allow single DN invoicing" checkbox in the delivery note type
- Make invoice button invisible if that setting is enabled


francesco-ooops (approved):
Functional ok

tafaRU (changes requested):
Grazie della PR!
Puoi documentare la nuova funzionalità nel README?

PR #4099, opened 2024-04-11
[14.0][IMP] l10n_it_delivery_note Grouping by product

toita86:
Currently the same product may appear on multiple lines within the same picking or across different pickings.

However, in the DDT (Delivery Document) printout, each product should appear only once (with the same batch, if activated).

Implemented:

Added a flag within the DDT type: "Consolidate lines per product in the report".

If this is activated, a DDT where the same product appears on multiple lines will only appear once in the printout.

If the "show order reference in DDT printout" setting is active and the grouped product comes from different orders merged into the same DDT, multiple orders would show in the "order ref" field on the printout.

The same is for the "show customer reference in DDT printout" setting.

### DIFF
--- a/l10n_it_delivery_note/models/stock_delivery_note.py
+++ b/l10n_it_delivery_note/models/stock_delivery_note.py
@@ -900,6 +900,74 @@ class StockDeliveryNote(models.Model):
 
         return location_address
 
+    def _grouped_lines(self):
+        def create_groups(show_so):
+            lines = self.env["stock.delivery.note.line"]
+            group_prod = self.line_ids.mapped("product_id")
+            for prod in group_prod:
+                filtered_lines = self.line_ids.filtered(lambda x: x.product_id == prod)
+                # for the case in which there are more discounts, price_unit, taxes
+                if len(set(filtered_lines.mapped("discount"))) > 1:
+                    lines += filtered_lines
+                    continue
+                if len(set(filtered_lines.mapped("price_unit"))) > 1:
+                    lines += filtered_lines
+                    continue
+                tax_ids = filtered_lines.mapped("tax_ids")
+                different_taxes = False
+                for line in filtered_lines:
+                    if not line.tax_ids == tax_ids:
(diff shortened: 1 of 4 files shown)


l10n_it_delivery_note/models/stock_delivery_note.py
```diff
@@ -900,6 +900,71 @@ def get_location_address(self, location_id):
...
+                different_taxes = False
+                for line in filtered_lines:
+                    if not line.tax_ids == tax_ids:
+                        lines += filtered_lines
+                        different_taxes = True
+                        break
+                if different_taxes:
+                    continue
+                if show_so:
+                    so_numbers = {line.sale_order_number for line in filtered_lines}
```
PicchiSeba, line 926: **issue**: if `show_so` is False the variable `so_numbers` is never declared, so its use at line 950 would raise an exception.

**suggestion**: we could replace these two lines with one-liner that also solves the issue above
```suggestion
                so_numbers = {line.sale_order_number for line in filtered_lines} if show_so else {}
```

PicchiSeba (changes requested):
Code Review:

l10n_it_delivery_note/models/stock_delivery_note.py
```diff
@@ -900,6 +900,71 @@ def get_location_address(self, location_id):
...
+                    uom = uom.search(
+                        [
+                            ("category_id", "=", uom[0].category_id.id),
+                            ("uom_type", "=", "reference"),
+                        ]
+                    )
+                line = self.env["stock.delivery.note.line"].new(
+                    {
+                        "product_id": prod,
+                        "name": prod.name,
```
aleuffre, line 940: **issue**: this one should be `filtered_lines[0].name`

aleuffre (changes requested):
Generally LGTM, a couple minor fixes (some have already been pointed out)

PicchiSeba (approved):
Code Review: LGTM

PR #4104, opened 2024-04-16
[14.0][IMP] l10n_it_fatturapa_out: use recipient code from commercial entity if not set

alessandro-fiorino:
If the recipient code is not set on the partner but it is on the commercial entity, use that instead of leaving it blank.


alessandro-fiorino:
Backport di https://github.com/OCA/l10n-italy/pull/4103 per la issue https://github.com/OCA/l10n-italy/issues/4102

PR #4112, opened 2024-04-22
[14.0] [FIX] l10n_it_delivery_note: report adjustments

renda-dev:
To display total prices a setting must be enabled and the DN's type must be "with prices"


francesco-ooops (approved):
LGTM

PR #4226, opened 2024-06-24
[14.0][FIX] l10n_it_delivery_note: unlink DNs from invoices at Cancel

toita86:
backport of https://github.com/OCA/l10n-italy/pull/4191
for the issue https://github.com/OCA/l10n-italy/issues/4190

PR #4232, opened 2024-06-25
[FIX] l10n_it_delivery_note: unlink DNs from invoices at Cancel

TheMule71:
Fixes: https://github.com/OCA/l10n-italy/issues/4190


TheMule71:
Chiudo in favore di https://github.com/OCA/l10n-italy/pull/4226
(non me n'ero accorto)

PR #4329, opened 2024-08-22
[16.0] [FIX] l10n_it_reverse_charge setting original invoice date in supplier self invoice

eLBati:
This way, VAT registry of supplier self invoices will contain the correct value for DATA FATT.

v14: https://github.com/OCA/l10n-italy/pull/4395


MaurizioPellegrinet (approved):
Test funzionale: OK

PR #4404, opened 2024-10-14
[16.0][FIX] p7m file encoded twice

### LINKED ISSUE #4405: [16.0] p7m file encoded twice by the supplier failed to be loaded
## Module
l10n_it_fatturapa / l10n_it_fatturapa_in

## Describe the bug
Somehow, supplier provide a p7m file that has been encoded twice in base 64. The module that allow us to load XML file decode the base 64 ony once and then the import failed.

## To Reproduce
Versions affected:
- [ ] `14.0`: yes
- [ ] `16.0`: https://github.com/OCA/l10n-italy/pull/4404

1. Encode any xml file twice with base 64
2. Try to import the Incoming E-bill Files

**Expected behavior**
File has to be decoded depiste the count of encoding

We have ask to our customer the authorization to share the with you.

yale-dynapps:
Sometimes suppliers encode the file twice. The first step to decode the file is not enough. Change the behaviour to be able to decode the file until it not coded in b64

### DIFF
--- a/l10n_it_fatturapa/__manifest__.py
+++ b/l10n_it_fatturapa/__manifest__.py
@@ -6,7 +6,7 @@
 
 {
     "name": "ITA - Fattura elettronica - Base",
-    "version": "16.0.1.3.0",
+    "version": "16.0.1.3.1",
     "category": "Localization/Italy",
     "summary": "Fatture elettroniche",
     "author": "Davide Corio, Agile Business Group, Innoviu, "
--- a/l10n_it_fatturapa/__manifest__.py
+++ b/l10n_it_fatturapa/__manifest__.py
@@ -6,7 +6,7 @@
 
 {
     "name": "ITA - Fattura elettronica - Base",
-    "version": "16.0.1.3.1",
+    "version": "16.0.1.3.0",
     "category": "Localization/Italy",
     "summary": "Fatture elettroniche",
     "author": "Davide Corio, Agile Business Group, Innoviu, "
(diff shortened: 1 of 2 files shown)


l10n_it_fatturapa/__manifest__.py
```diff
@@ -6,7 +6,7 @@
 
 {
     "name": "ITA - Fattura elettronica - Base",
-    "version": "16.0.1.3.0",
+    "version": "16.0.1.3.1",
```
francesco-ooops: Hi @yle-scopea , no need for this change, the version is bumped automatically on merge

francesco-ooops:
Hi @yle-scopea , thanks for your PR!

Could you create an issue (if possibile with a file to reproduce the problem) so we can track this problem across versions?

l10n_it_fatturapa/__manifest__.py
yale-dynapps: Ok I will avoid this next time

francesco-ooops: Nice, but you should avoid it this time as well 😅

yale-dynapps:
Hi @francesco-ooops 

I'm not able to provide you a file to test because it's my customer file not mine. Sorry.
But, the file is twice encoded in Base64. So first time, we get a new base64 raw data instead of XML readable file.

I will ask to my customer if I can share the file with you in an issue. Do  I need to create the issue even if I don't have file to provide?

Thanks

l10n_it_fatturapa/__manifest__.py
yale-dynapps: Done

francesco-ooops:
@yle-scopea yes please, open the issue as in this repo we try to track changes across versions and we can then link this PR to the issue

Do you know if there's a procedure to double encode the file? So we can use one for testing

sergiocorato (changes requested):
This change is dangerous, as data could be void and never return a value.

PR #4419, opened 2024-10-22
FIX l10n_it_account avoiding check_balance_sign_coherence for non IT countries

eLBati:
Otherwise, installing for instance French chart of account you would get something like "Incoherent balance signs for '42  Personnel et comptes rattachés' and its subgroups."


eLBati:
Sostituita da https://github.com/OCA/l10n-italy/pull/4431

PR #4581, opened 2025-01-28
[16.0] [FIX] l10n_it_withholding_tax: amount_net_pay_residual calculated incorrectly due to the mixed use of fields

eLBati:
Conditions: Invoice posted in USD currency for a company with the main currency in EUR. Invoice is partially reconciled in USD, one invoice line reconciled with 2 journal items:
- one journal items is exchange rates differences
- another item is partial payment

Basically, amount_net_pay_residual is calculated by taking USD amount and subtracting EUR amount, which leads to inconsistency

image


LorenzoC0 (approved):
LGTM 👍🏻

patrickt-oforce (approved):

PR #4606, opened 2025-02-06
[FIX] l10n_it_withholding_tax: fix amount_net_pay compute and add new field amount_net_pay_signed to show the value with a proper sign

lorenzoallegrucci:


matteoopenf:
quando puoi dai un occhio al pre-commit

PR #4634, opened 2025-03-07
[14.0][IMP] l10n_it_ricevute_bancarie: no delete for bank account used in riba

toita86:
solves  https://github.com/OCA/l10n-italy/issues/4618


SylweKra:
Superseeded by: https://github.com/OCA/l10n-italy/pull/4638

PR #4759, opened 2025-05-14
[FIX][l10n_it_account] Fix base amount tax partially undeductible in case of cash basis

alessandrocamilli:


andreampiovesana (approved):
LGTM

PR #4777, opened 2025-05-27
[14.0][FIX] l10n_it_fatturapa_in: fix monkey patch if field not exists

### LINKED ISSUE #4727: [l10n_it_fatturapa_in] Monkey patches cause errors when module is not installed
Hi,

## Module
l10n_it_fatturapa_in 

## Describe the bug
We run an Odoo 16.0 instance with multiple databases, some requiring Italian localization and others not. We keep OCA modules, including l10n-italy, in a shared addons_path.

After an update of the l10n-italy code in our addons_path, we noticed that databases where l10n_it_fatturapa_in is not installed can no longer create customer invoices (e.g., from Sales Orders).

The error seems related to code within l10n_it_fatturapa_in (we observed it originating from fields.py trying to access record.fatturapa_attachment_in_id). It appears this code runs and modifies Odoo's behavior as soon as it's loaded from the addons_path, even before the module is installed in a specific database. This causes an AttributeError because the required database fields are missing in databases where the module installation hasn't occurred.

## To Reproduce
**Affected versions**: 

- [x] 14.0: #4777 
- [ ] 16.0: sì
- [x] 18.0: No perchè non verrà portato a v18

1. install in the codebase the latest l10n-italy repo
2. access a database of the instance that do not use the module 
3. issue an invoice
4. observe the attribute error AttributeError: 'account.move.line' object has no attribute 'fatturapa_attachment_in_id'

**Expected behavior**
Issue an invoice normally even if the module is not installed

Thank you

GSLabIt:
fix issue  [4727](https://github.com/OCA/l10n-italy/issues/4727)

### DIFF
--- a/l10n_it_fatturapa_in/fields.py
+++ b/l10n_it_fatturapa_in/fields.py
@@ -24,7 +24,10 @@ def get_digits(self, env):
 
 
 def convert_to_cache(self, value, record, validate=True):
-    if record._name in ("account.move", "account.move.line"):
+    if (
+        record._name in ("account.move", "account.move.line")
+        and "fatturapa_attachment_in_id" in record._fields
+    ):
         e_invoice = record.fatturapa_attachment_in_id
         if e_invoice:
             # The invoice [line] has been created by importing an e-invoice.


francesco-ooops (approved):
Funzionale ok!

SicurSam (approved):

HekkiMelody (approved):

PR #4862, opened 2025-07-31
[FIX] l10n_it_reverse_charge : validating invoices without calling action_post does not generate self invoices

eLBati:
For instance, validating invoices massively from validate.account.move wizard


TheMule71:
Col merge di #4893, puoi fare un rebase per far girare i test.

PR #4867, opened 2025-08-01
[FIX]  l10n_it_account_vat_period_end_settlement: fix migration script

tafaRU:
According to
https://github.com/OCA/openupgradelib/blob/8a86014577d567883f73e36eea436af67096078d/openupgradelib/openupgrade.py#L2348-L2351 in order to have env as first parameter of migrate it needs the decorator openupgrade.migrate()


andreampiovesana (approved):
LGTM

PR #4963, opened 2025-10-27
[14.0] [FIX] l10n_it_account: fix script to add offset in printed page

### LINKED ISSUE #4964: [14.0] l10n_it_account: offset page not added in reports
## Module
l10n_it_account

## Describe the bug
Even if the start page number is set in wizard, that value is ignored

## To Reproduce
**Affected versions**:
- [ ] 14.0 - #4963 

1. Open accounting report wizard like vat registries or central journal
2. Set a start page value different from 0
3. Print the report
4. Check the result

**Expected behavior**
When a value is set for the start page field other than 0, that value must be added to current page number

patrickt-oforce:
Actually even the start page number is set, that value is not applied to printed pages; because the use of `eval` to get current page number.

Remove that call, that is not needed, solve the problem

### DIFF
--- a/l10n_it_account/reports/account_reports_view.xml
+++ b/l10n_it_account/reports/account_reports_view.xml
@@ -98,7 +98,7 @@
     </template>
 
     <!-- oca-hooks:disable=xml-dangerous-qweb-replace-low-priority -->
-    <template id="minimal_layout" inherit_id="web.minimal_layout">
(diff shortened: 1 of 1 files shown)


patrickt-oforce:
@OCA/local-italy-developers se ok, si può fare il merge?

andreampiovesana (approved):
LGTM

Issue #4969, opened 2025-11-04
[16.0] l10n_it_withholding_tax_financial_report: Module compliation issue observed


anusriNPS:
## Module
l10n_it_withholding_tax_financial_report

## Describe the bug
Compliation error observed due to recent OCA PR changes from "account_financial_reporting" module PR: #[1290](https://github.com/OCA/account-financial-reporting/pull/1290). Related xpath expression could not be found in parent view.

<img width="998" height="556" alt="Image" src="https://github.com/user-attachments/assets/f9d54398-e145-477c-a64a-d107b6d6b0bc" />

## To Reproduce
**Affected versions**:

1. Failure is observed when module is complied.

**Expected behavior**
Able to build module successfully.

- [x] 16.0 https://github.com/OCA/l10n-italy/pull/4971
- [x] 18.0 https://github.com/OCA/l10n-italy/pull/4976

PR #4976, opened 2025-11-06
[18.0][FIX] l10n_it_financial_statements_report: call super()

TheMule71:
After PR https://github.com/OCA/account-financial-reporting/pull/1408 it's better to use the provided _prepare_report_data() from the abstract model as it's pre-populated with needed data.

Solves: https://github.com/OCA/l10n-italy/issues/4969


MaurizioPellegrinet (approved):
Test funzionale: OK

PR #5097, opened 2026-02-06
[16.0][IMP] Add partner property_account_expense fallback in FatturaPA import

siddolo:
If module account_invoice_line_default_account is installed and the partner has a default expense account, use it as a fallback when there is no product match.


Borruso (approved):
LGTM

PR #5152, opened 2026-03-21
[14.0][FIX][l10n_it_withholding_tax] Fix amount wt with partial payme…

alessandrocamilli:
…nt. Logic aligned to v.16


monen17:
Hai rotto quasi tutto il rompibile :laughing: 
image
Puoi correggere?

PR #5180, opened 2026-04-17
[FIX] l10n_it_account: Updated bootstrap col

### LINKED ISSUE #5207: [FIX] l10n_it_account: wrong Bootstrap column classes in report header
## Module
l10n_it_account

## Describe the bug
The report header in `l10n_it_account/reports/account_reports_view.xml` uses Bootstrap 3 column classes (`col-xs-4`, `col-xs-offset-4`) which no longer exist in Bootstrap 5. This breaks the two-column layout in the header where company info and report title should sit side by side.

## To Reproduce
**Affected versions**: 18.00

1. Install `l10n_it_account`
2. Open any report that uses the header template
3. The header columns are not laid out correctly

**Expected behavior**
Company name/TIN on the left, report title on the right, properly spaced.

**Additional context**
Bootstrap 3 classes like `col-xs-*` were dropped in Bootstrap 4. The fix is replacing them with `col-6` and adding a small left margin to the right column.

- [18.0] https://github.com/OCA/l10n-italy/pull/5180

Matteo-Filardi:
Minor layout adjustment: updated column classes (col-xs-4 → col-6) and added margin to the right-side header column for better spacing. No functional changes.

### DIFF
--- a/l10n_it_account/reports/account_reports_view.xml
+++ b/l10n_it_account/reports/account_reports_view.xml
@@ -54,7 +54,7 @@
 
         <div class="header">
             <div class="row">
-                <div class="col-xs-4">
+                <div class="col-6">
                     <h3>
                     <span t-esc="company.name" />
                 </h3>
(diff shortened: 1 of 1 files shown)


Matteo-Filardi:
Fixes #5207

PR #5194, opened 2026-05-14
[18.0][FIX] l10n_it_vat_settlement_communication: fix for check_identificativo in multi-company

### LINKED ISSUE #5195: [18.0]l10n_it_vat_settlement_communication: "identifier already exists" error on save
## Module

  `l10n_it_vat_settlement_communication`

  ## Describe the bug

  Saving a `comunicazione.liquidazione` raises
  `ValidationError: Communication with identifier N already exists`
  even when no record with that identifier is visible to the user. The
  error fires on single-company databases and with a single company
  selected in the top-right switcher.

  Two defects combine.

  **1. `_get_identificativo` default**

  The default returns `count(records) + 1`. Row counting does not produce
  unique values. A deleted row, a gap, or an out-of-order import lets
  the default land on an identifier that already exists.

  **2. `_check_identificativo` constraint scope**

  The Python constraint searches without a `company_id` filter

  To Reproduce

  Affected versions: 18.0 (module 18.0.1.0.1)

  1. Create three comunicazione.liquidazione records on one company.
  The default assigns identifiers 1, 2, 3.
  2. Delete the record with identifier 2.
  3. Create a new record. The default returns 2 + 1 = 3, the save
  fails with
  ValidationError: Communication with identifier 3 already exists.

  Expected behavior

  The default returns a value that does not collide with an existing
  record. The uniqueness check matches records on the same company only.

- [x] 18.0 https://github.com/OCA/l10n-italy/pull/5194
- [ ] 19.0 https://github.com/OCA/l10n-italy/pull/5092

micheledic:
## Description

  The `_check_identificativo` Python constraint on `comunicazione.liquidazione`  ran `self.search([("identificativo", "=", self.identificativo)])` without a company filter. Creating a record on company A raises the duplicate error when company B already owns the same identifier because in the constrains defaults to allowed_company_ids instead of the current selected company.

  The `_get_identificativo` has the same problem, so two companies installed in the same database will return the first item also if more than one company is selected

  ## Steps to reproduce

  1. Install the module in a database with two companies, A and B.
  2. Log in as a user with access to both companies and the
     `account.group_account_manager` group.
  3. Create a `comunicazione.liquidazione` for company A. Note its
     `identificativo` (say, 1).
  4. Switch the active company to B and create a new
     `comunicazione.liquidazione`.
  5. The form picks the same `identificativo` and the Python constraint
     even though company B has none.

  ## Fix

  - Converted the `_check_identificativo` constraint.to  a SQL constraint `UNIQUE(company_id, identificativo)`. The database enforces uniqueness per company and ignores ORM record rules.
  - The `_get_identificativo` default by `company_id` from context or`env.company`, so the auto-assigned number counts records inside the current company.


micheledic:
@OCA/local-italy-maintainers rebased ed è green

tafaRU (approved):

PR #5202, opened 2026-05-28
[FIX] l10n_it_edi_extension: omit explicit ID in einvoice_line migration

tafaRU:
Explicitly passing the `id` column during the data migration from
`einvoice_line` to '`10n_it_edi_line` prevents the underlying PostgreSQL
sequence from advancing. As a result, the sequence lags behind the maximum
inserted ID, causing Odoo to crash with a "Duplicate key value violates
unique constraint" error on subsequent native record creations.

Fix this by omitting the `id` column from the INSERT and SELECT statements.
This allows PostgreSQL to handle ID generation natively via `nextval()`,
keeping the sequence and data naturally synchronized and avoiding the crash.

### DIFF
--- a/l10n_it_edi_extension/__init__.py
+++ b/l10n_it_edi_extension/__init__.py
@@ -714,12 +714,12 @@ def _l10n_it_fatturapa_in_post_migration(env):
                 """
                 INSERT INTO
                     l10n_it_edi_line (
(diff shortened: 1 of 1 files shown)


andreampiovesana (approved):
LGTM

sergiocorato:
Chiudo in quanto inclusa in #5203

PR #5219, opened 2026-06-09
[16.0] [IMP] l10n_it_account_stamp: avoid use distinct write to create account move line(s) for account stamp

### LINKED ISSUE #5220: l10n_it_account_stamp: write account stamp journal items result in unbalaced move
## Module
l10n_it_account_stamp

## Describe the bug
When confirm or reset to draft an invoice where need to apply account_stamp, an exception for unbalanced move is raised.

## To Reproduce
**Affected versions**:
- [x] 16.0 - #5219 
- [x]  18.0 - #5230 

1. Create an SO or an invoice with at least one line with tax Imponibile Escluso art. 15 (debito) and amount major than 77 €
2. Confirm the invoice
3.Receive Unbalanced move exception

**Expected behavior**
When the invoice is confirmed or reset to draft, this should happen without excpetion

**Additional context**
Actually cannot be reproduced in runboat, because l10n_it module (which contain italian accounting localization) cannot be installed, due to incompatibility between l10n_it_fatturapa and l10n_it_edi which is automatically installed when l10n_it is installed

patrickt-oforce:
Sometimes can happen when reset to draft an invoice or confirm it, that operation raise an exception for unbalanced move. Due to creation of account stamp move line(s) is executed in two different step even if context key `check_move_validity` is present with value `False`.

This PR, execute the create command in a single step, avoid calling multiple times methods to recompute dynamic lines

### DIFF
--- a/l10n_it_account_stamp/models/account_move.py
+++ b/l10n_it_account_stamp/models/account_move.py
@@ -3,6 +3,7 @@
 
 from odoo import _, api, fields, models
 from odoo.exceptions import UserError
+from odoo.fields import Command
 
 
 class AccountMove(models.Model):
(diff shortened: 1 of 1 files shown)


MaurizioPellegrinet (approved):
Test funzionale: OK

tafaRU (approved):

PR #5228, opened 2026-06-16
[16.0] [IMP] l10n_it_withholding_tax: pass default name to account move for withholding tax

### LINKED ISSUE #5229: l10n_it_withholding_tax: not set name in withholding tax payment move vals raise duplicate account.move exception
## Module
l10n_it_withholding_tax

## Describe the bug
Sometime can happen that when register a payment for a supplier invoice subjected to withholding tax, during the account.move payment confirmation duplicate account.move exception is raised

## To Reproduce
**Affected versions**:
- [ ] 16.0 - #5228 
- [ ] 18.0 - ?

1. Configure withholding tax
2. Create a supplier invoice with at least one line with withholding tax
3. Confim the invoice
4. Register payment; if this is the first invoice subjected to withholding tax, the error can't happen; because there no other payment account move for withholding tax. In that case, the error happen from second payment

**Expected behavior**
When a payment for invoice subject to withholding tax is registered, that payment can be confirmed

patrickt-oforce:
Sometime can happen that not passing name value in Withholding Tax payment account move, lead to a duplicate sequence error. 
This because, the account.move name field is used here: https://github.com/odoo/odoo/blob/16.0/addons/account/models/sequence_mixin.py#L76
and during the confirmation of Withholding tax payment account.move, the sequence is set to restart from 0 and this sequence restart raising a duplicate account move name

### DIFF
--- a/l10n_it_withholding_tax/models/withholding_tax.py
+++ b/l10n_it_withholding_tax/models/withholding_tax.py
@@ -370,7 +370,12 @@ class WithholdingTaxMove(models.Model):
                 % (self.wt_account_move_id.name)
             )
         # Move - head
+        # Add name key in move vals, avoid to short circuit sequence date
+        # computation in
+        # https://github.com/odoo/odoo/blob/16.0/addons/account/models/sequence_mixin.py#l76
+        # that short circuit lead to duplicate account move exception
         move_vals = {
+            "name": "/",
             "ref": _(
                 "WT %(code)s - %(move)s",
                 code=self.withholding_tax_id.code,


patrickt-oforce:
@OCA/local-italy-developers si può fare il merge?

francesco-ooops:
C'è qualche volontario per una review funzionale?

monen17 (changes requested):
I just tried in runboat and there is no issue, see https://github.com/OCA/l10n-italy/issues/5229#issuecomment-5371693151.

PR #5259, opened 2026-07-27
[14.0] [FIX] l10n_it_fatturapa_in: convert received date to context timezone

### LINKED ISSUE #2778: l10n_it_fatturapa_in, Setting Italian timezone by default, otherwise e_invoice_received_date could be set the day before of the actual date
## To Reproduce
**Affected versions**:

 - [ ] 12.0: https://github.com/OCA/l10n-italy/pull/2777
 - [ ] 14.0

receiving date = 2022-04-01 00:16
saved in DB, UTC format, as 2022-03-31 22:16:03
converted to date = 2022-03-31

In this case, the invoice would be registered in the wrong period (month)

eLBati:
The e-invoice received date was extracted with `received_date.date()` directly on `e_invoice_received_date` (a Datetime field, stored naive in UTC). Taking `.date()` in UTC returns the wrong day for invoices received in the early hours of the local morning, which in UTC still fall on the previous day.

Example: an invoice received at 01:18 CEST (UTC+2) is stored as 23:18 UTC of the previous day, so `.date()` yielded the day before. This wrong date then propagated to the accounting registration date (`date`) when the company uses `in_invoice_registration_date == "rec_date"`.

Convert the value with `fields.Datetime.context_timestamp` before calling `.date()`, so the reception (and registration) date matches the local timezone.

Assisted-by: Claude Opus 4.8

### DIFF
--- a/l10n_it_fatturapa_in/wizard/wizard_import_fatturapa.py
+++ b/l10n_it_fatturapa_in/wizard/wizard_import_fatturapa.py
@@ -1121,7 +1121,13 @@ class WizardImportFatturapa(models.TransientModel):
         received_date = attachment.e_invoice_received_date
         if not received_date:
             received_date = attachment.create_date
-        received_date = received_date.date()
+        # received_date is a naive UTC datetime; convert it to the context
+        # timezone before extracting the date, otherwise invoices received in
+        # the early hours (still the previous day in UTC) get a reception date
+        # (and accounting registration date) one day earlier than the local one.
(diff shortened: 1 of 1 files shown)


francesco-ooops:
@eLBati risolve questa vero? https://github.com/OCA/l10n-italy/issues/2778

eLBati:
> @eLBati risolve questa vero? #2778

Ah, sì, grazie

PR #5297, opened 2026-09-02
[18.0][FIX] l10n_it_edi_doi_extension: wrong sign on sale refunds in _compute_invoiced and other fixes

micheledic:
### Description 
A customer credit note under a declaration of intent was raising the plafond instead of lowering it. Core already applies -direction_sign to l10n_it_edi_doi_amount, so an out_refund comes in negative and _compute_invoiced negated it a second time. Bill amounts are computed by this module without sign, so the negation stays for in_refund only.

### Fixes
- Apply the absolute value so it is not double negative the value on refund
- Append the declaration of intent warning instead of overriding it
- The _reverse_moves override is replaced by copy_data adding the bridge lines when reversed_entry_id is in default: core _reverse_moves goes through copy(), so the lines are created while the move is still draft, while the override recreated them after super(), on an already posted move when cancel=True. Duplicates keep following core (no bridge lines): the split between declarations depends on their remaining amount at invoicing time.
- raise error if amount <= 0 , before the helps sait that if amount=0 the full amount is used but never implemented in the code 
- If it is a DOI migrated from 16.0 now it used the amount_untaxed as fallback written by post_init_hook
- Use price_subtotal instead of price_total so if there are more than one taxes, only the subtotal of doi tax is used . For example one line two different taxes , one with tax_amount > 0 , the tax_amount was added to the DOI amount. It is wrong
- Make the bridge DOI lines readonly if posted because deleting them on posted invoice was clearing l10n_it_edi_doi_id , if it was already sent the data in XML becames different than the data on invoice
- Add permission on group_account_readonly group

### DIFF
--- a/l10n_it_edi_doi_extension/models/account_move.py
+++ b/l10n_it_edi_doi_extension/models/account_move.py
@@ -1,6 +1,6 @@
 # Copyright 2025 Nextev Srl
 
-from odoo import _, api, fields, models
+from odoo import Command, _, api, fields, models
 from odoo.exceptions import UserError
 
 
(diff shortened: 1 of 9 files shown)


micheledic:
@OCA/local-italy-maintainers rebased

PR #5300, opened 2026-09-04
[18.0][FIX] l10n_it_asset_management: monthly depreciation pro rata

### LINKED ISSUE #5301: [18.0] l10n_it_asset_management: monthly depreciation with pro rata is about 12 times too low
## Module

`l10n_it_asset_management`

## Describe the bug

When the "Generate Depreciations" wizard runs with period "Month" on an asset whose depreciation has "Pro-rata Temporis" checked, the quota comes out around twelve times too low.

`get_depreciation_amount_multiplier` divides the multiplier by 12 for the month, then multiplies it again by the days elapsed over the whole fiscal year when pro rata is active. The month is already a twelfth of the year, so the day fraction cuts the same amount a second time. December is the only month that comes out right, because its fraction of days reaches 1.

The previsional report is not affected: it calls `generate_depreciation_lines` without a period, so it takes the yearly path.

## To Reproduce

1. Create an asset of 15.000 with purchase date 16/07/2026.
2. On its civil depreciation set 20% and check "Pro-rata Temporis".
3. Run "Generate Depreciations" with date 31/07/2026, period "Month", 1 period.
4. Repeat for 31/08/2026.

## Current behavior

| Depreciation date | Amount |
|---|---|
| 31/07/2026 | 10,96 |
| 31/08/2026 | 32,19 |
| 31/01/2027 | 21,23 |

## Expected behavior

The full monthly quota is 15.000 x 20% / 12 = 250. July should count the 16 days of ownership out of 31, so 129,03, and every later month should get 250.

## Additional context

Quotas already generated keep their wrong amount: the wizard does not recompute existing lines.
- [16.0]
- [18.0] #5300

micheledic:
Monthly depreciation on a pro rata asset comes out about twelve times too low. A customer got 10,96 for July where the quota is 250 (asset of 15.000 at 20%, registered 16/07).

The multiplier gets divided by 12 for the month, then multiplied by the days elapsed since the start of the fiscal year. The month is already a twelfth of the year, so the day fraction cuts it twice. Only December survives, its fraction being 1.

The monthly branch now looks at the month instead of the year: 16 days out of 31 in July, full quota from August. Yearly depreciations keep the old path, and so does the previsional report, which passes no period.

New test: 2,58 / 5 / 10 with the fix, 0,22 / 0,64 / 2,96 without.

### DIFF
--- a/l10n_it_asset_management/models/asset_depreciation.py
+++ b/l10n_it_asset_management/models/asset_depreciation.py
@@ -7,6 +7,7 @@ from odoo import api, fields, models
 from odoo.exceptions import ValidationError
 from odoo.fields import Command
 from odoo.tools import float_compare, float_is_zero
+from odoo.tools.date_utils import relativedelta
 
 
 class AssetDepreciation(models.Model):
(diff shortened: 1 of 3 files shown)


micheledic:
@OCA/local-italy-maintainers rebased

PR #5307, opened 2026-09-11
[18.0] [FIX] l10n_it_riba_oca: post the credit entry, allow to cancel slips

eLBati:
The credit entry was created in draft, so its line on the RiBa account could not be reconciled until the user posted it by hand, while both the settlement and the past due reconcile it. The wizard posts it now.

Cancelling an accepted slip failed, because its acceptance entries are posted and reconciled with the invoices:

    You cannot do this modification on a reconciled journal entry.

The entries of the slip are now set back to draft, which also removes their reconciliation, before being deleted.

### DIFF
--- a/l10n_it_riba_oca/models/riba.py
+++ b/l10n_it_riba_oca/models/riba.py
@@ -174,16 +174,26 @@ class RibaList(models.Model):
             for line in slip.line_ids:
                 line.confirm()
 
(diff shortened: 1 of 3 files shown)


eLBati:
Includo in https://github.com/OCA/l10n-italy/pull/5308

